PR #51, opened 2025-07-14
docs: add cli generator

ArthurFlag:


crazy-max (changes requested):
You also need to add a job to validate docs similar to https://github.com/docker/model-cli/blob/main/.github/workflows/validate.yml

ArthurFlag:
@crazy-max, David kindly fixed the problem, things are working now, would you mind taking a final look when you can?

.dockerignore
```diff
@@ -4,4 +4,5 @@
 !/go.mod
 !/go.sum
 !/.golangci.yml
-!/vendor
\ No newline at end of file
+!/vendor
+!/docs
```
crazy-max, line 8: Why? This would make the validation failed

crazy-max (changes requested):
> You also need to add a job to validate docs similar to https://github.com/docker/model-cli/blob/main/.github/workflows/validate.yml

@ArthurFlag ^

See also my previous review for the Dockerfile

.dockerignore
ArthurFlag: That's an exception so we're good 👍  (ignore everything but /docs)

crazy-max: Oh indeed my bad!

crazy-max (changes requested):
Small nit otherwise nicely done!

Also I guess as follow-up there will be some examples and a better description for commands. Atm it's hard to understand what they mean and what they do if we publish this on docs website.

crazy-max (changes requested):
:eyes: https://github.com/docker/mcp-gateway/actions/runs/16490847205/job/46624943785?pr=51#step:3:479

```
#15 0.694 ERROR: Docs result differs. Please update with "make docs"
#15 0.708  M docs/generator/reference/docker_mcp_gateway_run.yaml
```

ArthurFlag:
mmmh, I can't reproduce locally, I'm not sure how to diagnose further, any idea?

crazy-max:
That's the because validation is based on merged pull request with base branch so you need to rebase and run `make docs` again. This is to avoid merging faulty changes.

I just tried on my side and it works. You can cherry-pick this commit if you want: https://github.com/crazy-max/docker-mcp-gateway/commit/1d676e1a4e2a840010255ea00bcea3e183bde483

Issue #48, opened 2025-07-11
Support Goose


cmrigney:
`docker mcp client` connect should support Goose: https://block.github.io/goose/docs/category/getting-started/

cmrigney:
cc @Daniel-Kolev

PR #53, opened 2025-07-15
Add goose client

Daniel-Kolev:
**What I did**

Add Goose as a global client - https://github.com/block/goose

It's a bit different than the other global clients - the gateway is specified with a whole new yaml config. The name also can't contain UPPER CASE characters or some special characters ([their app removes those](https://github.com/block/goose/blob/c6225e38d02295ef65714c1e87ff6494ac0f8b04/ui/desktop/src/extensions.tsx#L62-L64)) hence the `SIMPLE_NAME` change I made. 

Resulting config in the goose config.yaml:
```yaml
  mcpdocker:
    args:
      - mcp
      - gateway
      - run
    bundled: null
    cmd: docker
    description: The Docker MCP Toolkit allows for easy configuration and consumption of MCP servers from the Docker MCP Catalog
    enabled: true
    env_keys: []
    envs: {}
    name: mcpdocker
    timeout: 300
    type: stdio
```
As seen in their app:
image

The bundled extensions in Goose respect the displayName field but that doesn't seem to be the case for custom ones so we are stuck with `Mcpdocker` 😞 

**Related issue**

fixes https://github.com/docker/mcp-gateway/issues/48


cmd/docker-mcp/client/config.yml
```diff
@@ -41,6 +41,31 @@ system:
...
+  goose:
+    displayName: Goose
+    source: https://github.com/block/goose
+    installCheckPaths:
+    - $HOME/.config/goose
+    - $USERPROFILE\.config\goose
+    paths:
+      linux: $HOME/.config/goose/config.yaml
+      darwin: $HOME/.config/goose/config.yaml
+      windows: $USERPROFILE\.config\goose\config.yaml
```
Daniel-Kolev, line 53: I haven't checked this myself

ipedrazas (approved):

dgageot (approved):

Daniel-Kolev:
Will merge after I get https://github.com/docker/mcp-gateway/pull/55 in first and add the goose icon here

Daniel-Kolev:
We can merge and release this PR but we shouldn't push it to Docker Desktop 4.43.x or earlier because Docker Desktop wouldn't show the proper icon. With 4.44 and later, it will (https://github.com/docker/pinata/pull/35604)

Daniel-Kolev:
Talked with @nunocoracao and he is fine with releasing this even with a ? icon for 4.43.x

PR #62, opened 2025-07-18
Tools filtering

Daniel-Kolev:
**What I did**

Tool filtering

Add new commands: `docker mcp tools enable <toolName>` and `docker mcp tools disable <toolName>`. They auto-discover the first matching server using the catalog. Also they accept a `--server` flag. It helps with duplicate tool names where the auto-discover might fail to find the server the user wants.

The enabled tools are stored in a `tools.yaml` file which contains all the enabled tools grouped by server. Example file:
```
curl: []
filesystem:
  - directory_tree
  - get_file_info
  - list_allowed_directories
  - list_directory
  - read_file
  - read_multiple_files
  - search_files
```
read as:
```
curl has all of its tools disabled
filesystem has some of its tools enabled
if a server is not mentioned in the file, then all of its tools are enabled
```

The gateway reads and respects this file. When launching the gateway, the existing `--tools` option (which accepts the tools names) takes precedence over the `tools.yaml` file. This works the same way as the `--servers` option and the `registry.yaml`.

Nice to haves: remove the server if all of its tools are enabled

**Related issue**

### DIFF
--- a/cmd/docker-mcp/tools/enable.go
+++ b/cmd/docker-mcp/tools/enable.go
@@ -1,0 +1,152 @@
+package tools
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"sort"
+
+	"gopkg.in/yaml.v3"
+
+	"slices"
+
+	"github.com/docker/mcp-gateway/cmd/docker-mcp/internal/catalog"
+	"github.com/docker/mcp-gateway/cmd/docker-mcp/internal/config"
+	"github.com/docker/mcp-gateway/cmd/docker-mcp/internal/docker"
+)
+
+func Disable(ctx context.Context, docker docker.Client, toolNames []string, serverName string) error {
+	return update(ctx, docker, nil, toolNames, serverName)
+}
+
+func Enable(ctx context.Context, docker docker.Client, toolNames []string, serverName string) error {
+	return update(ctx, docker, toolNames, nil, serverName)
+}
+
+func findServerByTool(mcpCatalog catalog.Catalog, toolName string) (string, error) {
+	for serverName, server := range mcpCatalog.Servers {
+		for _, tool := range server.Tools {
+			if tool.Name == toolName {
+				return serverName, nil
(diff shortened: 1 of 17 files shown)


cmd/docker-mcp/tools/enable.go
Daniel-Kolev, line 1: I could probably give this file a bit more love

dgageot:
Hey @cmrigney, I didn't have time to look at Daniel's PR. Could you? Thanks!

cmd/docker-mcp/tools/tools_test.go
cmrigney, line 1: Nice tests! 🎉

cmd/docker-mcp/tools/tools_test.go
```diff
@@ -0,0 +1,320 @@
...
+      - name: "search_duckduckgo"
+        description: "Search DuckDuckGo"
+      - name: "other_tool"
+        description: "Another tool"
+  other_server:
+    tools:
+      - name: "other_tool"
+        description: "Another tool"
+`
+		writeFile(t, filepath.Join(home, ".docker/mcp/catalogs/docker-mcp.yaml"), []byte(catalogContent))
```
cmrigney, line 318: **suggestion**: I think if I run these tests in VSCode instead of `make test` (in a container), it will break my local setup right? In that case, it might be a good idea to read the existing file and add a `t.Cleanup` that re-writes it back to the original.

cmrigney (approved):
Nice addition! lgtm, just left a couple optional suggestions.

cmd/docker-mcp/tools/tools_test.go
Daniel-Kolev: There's a section in the setup() function that creates a temp directory and replaces the home one with that new temporary one. It gets cleaned up after the tests run so it actually doesn't interfere with your local setup. I did run the test locally and my tools.yaml didn't change.

image

Issue #60, opened 2025-07-18
Prevent starting in stdio mode when `--port` is specified


eharris128:
# Current state
User is able to attempt to run the gateway in a way where stdio is attempted to be served over http:

<img width="1123" height="578" alt="Image" src="https://github.com/user-attachments/assets/7e191187-a239-4476-bea3-180952a79d98" />

## Observation
- This does not work. Crashing out sooner, or preventing the pairing of these modes would be better.

slimslenderslacks:
@eharris128 the combination of a `--port` and stdio activates a mode where multiple clients can connect multiple stdio clients with tools like socat

```
socat STDIO TCP:localhost:8080
```

We found this useful for starting up a long running gateway process but still permitting multiple stdio clients to use it.  Now that we support other transports like sse and streaming, and clients are beginning to support this, this will probably become much less useful.

We could potentially add another option to `--transport` that would force users to acknowledge that we're starting up a listening tcpip port.  This would allow us to error out with with the set of options you specified above.

@dgageot wdyt? We could think about making `--transport=tcp --port=8080` starting up the socket stdio servers, and have default stdio transport with a port become an error.

eharris128:
Learning moment for me - ty.

My observation feels a little like beginner's ignorance now ;)

dgageot:
This is fixed now.

eharris128:
Many thanks again.

PR #73, opened 2025-07-22
Deprecate stdio over TCP

dgageot:
**What I did**

Remove the stdio over TCP mode. It's more of a hack, has a couple of nasty issues and makes the code more complicated.

**Related issue**

https://github.com/docker/mcp-gateway/issues/60

### DIFF
--- a/cmd/docker-mcp/internal/gateway/run.go
+++ b/cmd/docker-mcp/internal/gateway/run.go
@@ -6,7 +6,6 @@ import (
 	"net"
 	"os"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/mark3labs/mcp-go/server"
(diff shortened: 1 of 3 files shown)


cmd/docker-mcp/internal/gateway/run.go
```diff
@@ -200,19 +164,20 @@ func (g *Gateway) listServersAndCapabilities(ctx context.Context, configuration
...
-func refreshCapabilities(s *server.MCPServer, c *Capabilities) {
-	s.SetTools(c.Tools...)
-	s.SetPrompts(c.Prompts...)
-	s.SetResources(c.Resources...)
-	s.RemoveAllResourceTemplates()
-	for _, v := range c.ResourceTemplates {
-		s.AddResourceTemplate(v.ResourceTemplate, v.Handler)
+	// Update the server's capabilities.
+	g.health.SetUnhealthy()
+	mcpServer.SetTools(capabilities.Tools...)
```
slimslenderslacks, line 173: I'm actually kind of worried about the conflicts here because we could end up with conflicting tools appearing and disappearing depending on which server last changed.

slimslenderslacks (approved):
Understood and looks good.
We should talk about managing conflicts separately.

Issue #15, opened 2025-06-25
Eliciation Support (version 2025-06-18)


slimslenderslacks:
## Background

We need to plan when to update our server capabilities to 2025-06-18.  However, when we have done so, note that [eliciation support](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/382) is now merged into that version of the specification.  

### How will we see this

While processing a tool request, or for [long running](#5) MCP servers that wake up and need information from a client, an MCP server may put a client `elicitation/request` on the outbound channel. See above pull request for sequence diagrams.

slimslenderslacks:
This is supported since version 0.15.

When a server creates an elicitation, it will be forward through the gateway to the client.

PR #77, opened 2025-07-29
Switch to modelcontextprotocol/go-sdk

### LINKED ISSUE #79: MCP server roots/list requests are not being delivered to the client.
## Background

In the current gateway, when a MCP Server running behind the gateway makes a roots/list request, it is not being delivered to any of the current client sessions.  This means that if a server is running behind the gateway it can not ask attached clients for their current Roots.

slimslenderslacks:
**What I did**

Refactor gateway to use the official GO SDK

**Related issue**

* Prepares for #15 (elicitation support).
* Fixes #79 - missing tool list change notifications

**Changes**

* switches custom mcp client for official client SDK (major change)
    * this is a major change because our custom client was fairly minimal and tended to not be less strict with tool call payloads
* server sdk is switched from mark3labs to the official SDK (minor chanage)


cmd/docker-mcp/internal/mcp/remote.go
```diff
@@ -63,35 +56,38 @@ func (c *remoteMCPClient) Initialize(ctx context.Context, request mcp.Initialize
...
-			return nil, err
-		}
+		// TODO: Need to implement custom HTTP client with headers for SSE
+		mcpTransport = mcp.NewSSEClientTransport(url, &mcp.SSEClientTransportOptions{})
 	case "http", "streamable", "streaming", "streamable-http":
-		remoteClient, err = client.NewStreamableHttpClient(url, mcptransport.WithHTTPHeaders(headers))
-		if err != nil {
-			return nil, err
-		}
+		// TODO: Need to implement custom HTTP client with headers for streaming
```
cmrigney, line 67: **question:** What's this needed for?

cmrigney (approved):
Nice work! There could be one blocking issue, see my comments. Also, I am still testing the changes out. But code lgtm in general 👍

cmd/docker-mcp/internal/mcp/remote.go
slimslenderslacks: I don't think we need these TODOs any longer.  This requriement has been absorbed into the go-sdk

PR #98, opened 2025-08-20
Expose enabled flag for all tools

Daniel-Kolev:
**What I did**

Building on top of the changes I did in https://github.com/docker/mcp-gateway/pull/62 now I am exposing an "enabled" flag for each tools so I could consume it in Docker Desktop and allow uses to enable/disable specific tools.

**Related issue**


masegraye (approved):
Good.

PR #110, opened 2025-08-26
Fix parsing bug that breaks Docker Desktop client listing

cmrigney:
**What I did**
Docker Desktop's MCP client listing UI was broken because of spurious output in the `docker mcp client ls --global` command when a `type` wasn't recognized.

For example, if the Claude config contained an `http` type like:
```
{
  "mcpServers": {
    "my-server": {
      "type": "http",
      "url": "https://docker.com"
    }
  }
}
```

Then `docker mcp client ls --global` would produce:
```
unknown server type for "http"
{
<rest of the json>...
```

I fixed it by printing the warning on stderr instead of stdout. While I was in there, I also added support for printing `http` configurations in the client listing.

**Related issue**

https://github.com/docker/for-win/issues/14908

### DIFF
--- a/cmd/docker-mcp/client/parse.go
+++ b/cmd/docker-mcp/client/parse.go
@@ -3,6 +3,7 @@ package client
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"strings"
 )
 
@@ -14,6 +15,7 @@ import (
 type MCPJSONLists struct {
 	STDIOServers []MCPServerSTDIO
 	SSEServers   []MCPServerSSE
+	HTTPServers  []MCPServerHTTP
 }
 
 type MCPServerSTDIO struct {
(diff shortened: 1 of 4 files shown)


cmd/docker-mcp/client/parse.go
```diff
@@ -72,8 +85,14 @@ func UnmarshalMCPJSONList(data []byte) (*MCPJSONLists, error) {
...
 			cfg.SSEServers = append(cfg.SSEServers, server)
+		case "http":
+			var server MCPServerHTTP
+			if err := json.Unmarshal(raw, &server); err != nil {
+				return nil, err
+			}
+			cfg.HTTPServers = append(cfg.HTTPServers, server)
 		default:
-			fmt.Printf("unknown server type for %q\n", itemType)
+			fmt.Fprintf(os.Stderr, "unknown server type for %q\n", itemType)
```
cmrigney, line 95: This was the bug fix.

PR #121, opened 2025-09-03
Show error when OAuth provider doesn't exist

kgprs:
**What I did**
If the user types in `docker mcp oauth authorize something_that_doesnt_exist`, it prints "Opening your browser for authentication. If it doesn't open automatically, please visit" instead of "OAuth provider does not exist".

**Tested**
OAuth Provider exists:
```
./docker-mcp oauth authorize github
Opening your browser for authentication. If it doesn't open automatically, please visit: <link>
```

OAuth Provider doesn't exist:
```
./docker-mcp oauth authorize github2
OAuth provider does not exist
```

### DIFF
--- a/cmd/docker-mcp/internal/desktop/raw_client.go
+++ b/cmd/docker-mcp/internal/desktop/raw_client.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"fmt"
 	"io"
 	"net"
 	"net/http"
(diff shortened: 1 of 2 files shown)


slimslenderslacks (approved):
lgtm

kgprs:
> lgtm - one enhancement we can make is to only allow this for the providers which show up in: `docker mcp oauth ls`

Good call - that'll find the issue and exit earlier. I'll leave this in for now and we can optimize it later

PR #124, opened 2025-09-05
Update Windows location of Goose config.yaml file

slimslenderslacks:
* see [doc](https://block.github.io/goose/docs/guides/config-file/)


cmrigney (approved):
Verified it works on windows!

PR #128, opened 2025-09-08
Show `server list` command

bobbyhouse:
Remove the hidden flag from the server list command

**What I did**
Removed `Hidden: true` so that `docker mcp server` would show documentation for `docker mcp server list`

**Related issue**
None

### DIFF
--- a/cmd/docker-mcp/commands/server.go
+++ b/cmd/docker-mcp/commands/server.go
@@ -45,7 +45,6 @@ func serverCommand(docker docker.Client) *cobra.Command {
 
 			return nil
 		},
-		Hidden: true,
 	}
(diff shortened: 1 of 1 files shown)


slimslenderslacks:
@masegraye @ipedrazas can either of you review this one? I'm not really sure why it's hidden and I think it makes sense to have it. My only concern is that the `docker mcp server ...` set of commands conflicts with working sets.

slimslenderslacks (approved):
let's merge this

slimslenderslacks:
@bobbyhouse I kind of wanted to discuss this with the team but I think you should merge it now.  It makes everything more consistent.

PR #132, opened 2025-09-10
Add telemetry for initialize

cmrigney:
**What I did**
Added telemetry for initialize when a client connects to the gateway. It includes the client and version.

**Related issue**


slimslenderslacks (changes requested):
This looks totally right to me.  Is there anyway you can add a test to go along with other telemetry tests?

PR #147, opened 2025-09-17
Make it easier to use docker-mcp as a library

dgageot:
**What I did**

Rename the pesky `internal` package into `pkg`.

This makes it much easier to vendor small portions of the gateway into other products and use the MCP gateway as a library.


slimslenderslacks:
I moved this to pkg in the root of the project.

Issue #160, opened 2025-09-27
Documentation update for connection to OpenAI Codex


cpjet64:
Here are working examples of different MCP Client connection configs for OpenAI Codex:

Windows 11 Docker Desktop (Codex running in CMD or Powershell)
[mcp_servers.docker-mcp-gateway]
args = ["mcp", "gateway", "run"]
command = "docker.exe"
startup_timeout_ms = 10000

Windows 11 Docker Desktop (Codex running in WSL Distro such as Ubuntu)
[mcp_servers.docker-gateway]
command = "/mnt/c/Program Files/Docker/Docker/resources/bin/docker.exe"
args = ["mcp", "gateway", "run"]

slimslenderslacks:
@cpjet64 thank you for this.

Perhaps we can also use this to track an update to the `docker mcp client connect codex` command to make this update. At the moment, I think we're just looking for a round-tripping toml parser.

slimslenderslacks:
Thanks, that would be amazing. For the json edits, we've been using `github.com/mikefarah/yq/v4 v4.45.4` to make edits, but I think that has limited support for editing toml files. Ideally, we can:

1. list each `mcp_server` and check for ones with a particular name (eg {our_name})
2. add missing `mcp_server` entries
3. delete the `mcp_server.{our_name}` entry if the user disconnects

And make sure that we preserve all comments and formatting for any edits that we make.

To be honest, if you just ask codex to edit it's own file, it does a fine job, so it feels kind of odd that we're trying to find a good toml parser :)

slimslenderslacks:
Oh wow, great find @cpjet64.

I've started a PR #165 to integrate this.  So we'll use the existence of the command on the current shell to determine whether or not it's installed and then I've built up from there.

cpjet64:
Glad to of been of service!

slimslenderslacks:
It almost seems impossible to use codex from WSL given how much enviroment you have to copy in.

PR #165, opened 2025-09-30
Add OpenAI codex client support

slimslenderslacks:
**What I did**

Add support for OpenAI codex in `docker mcp client ...`

**Related issue**

#160

### DIFF
--- a/cmd/docker-mcp/client/codex_handler.go
+++ b/cmd/docker-mcp/client/codex_handler.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 
 	"github.com/pelletier/go-toml/v2"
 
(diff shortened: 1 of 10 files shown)


cmd/docker-mcp/client/codex_handler.go
```diff
@@ -0,0 +1,59 @@
...
+	"context"
+	"os/exec"
+	"strings"
+)
+
+// isCodexInstalled checks if the codex binary is installed and working
+func isCodexInstalled(ctx context.Context) bool {
+	err := exec.CommandContext(ctx, "codex", "--version").Run()
+	return err == nil
+}
```
xenoscopic, line 13: If we try to drive `codex` from Docker Desktop (including via a `docker mcp` subcommand), there may be a high probability of it not being in the Docker Desktop process' `PATH` (esp. if the user has installed it with `nvm`).  I don't know if there's a good solution to that.

cpjet64: One thing I will warn you all about is be ready for the support questions if Codex takes their time (like they have been doing) with Windows bugfixes. They still havent fixed the PATH/PATHEX issues yet so users still need to use the environment variables I specified in #160. 

It doesnt hurt to use the full command its just unnecesary once the bugfix is implemented. Also the full command is not needed at all for the WSL implementation.

slimslenderslacks: @Daniel-Kolev what about if we backup with the config directory check.  And then for the connect/disconnect, where we want to use `codex` because we're worried about not having a good toml parser, what kind of errors can we provide to the DD gui (if codex is not available in the path)?

PR #176, opened 2025-10-08
Send OAuth URL as plaintext in mcp-add's response when elicitations are not supported

kgprs:
https://docker.atlassian.net/browse/MCP-125

Most clients don't support elicitations yet, so this will send back the login link as text so that these clients can ask the user to login. Elicitation support added here: https://github.com/docker/mcp-gateway/pull/175

Tested with goose:
Screenshot 2025-10-07 at 10 12 58 PM


slimslenderslacks (approved):
looks good!

PR #190, opened 2025-10-19
Add authentication for SSE and streaming gateway modes

slimslenderslacks:
Implement secure authentication for HTTP-based gateway transports (SSE and streaming) to protect MCP server access. The authentication system supports two methods: query parameter tokens and HTTP Basic Auth.

Key features:
- Auto-generates 50-character cryptographically secure tokens using lowercase letters and numbers
- Respects MCP_GATEWAY_AUTH_TOKEN environment variable for custom tokens
- Supports query parameter auth via ?MCP_GATEWAY_AUTH_TOKEN=<token>
- Supports HTTP Basic Auth with any username and token as password
- Health endpoint (/health) remains accessible without authentication
- Uses constant-time comparison to prevent timing attacks
- Outputs complete authenticated URLs on gateway startup

The gateway now displays the full URL with embedded token on startup, making it easy to copy and use in client configurations. Basic Auth credentials are also shown as an alternative authentication method.

This change prevents DNS rebinding attacks by forcing any local gateway runtime to run with an auth token, ensuring that malicious websites cannot exploit the locally-running gateway service.

### DIFF
--- a/docs/generator/reference/docker_mcp_gateway_run.yaml
+++ b/docs/generator/reference/docker_mcp_gateway_run.yaml
@@ -275,7 +275,8 @@ options:
     - option: transport
       value_type: string
       default_value: stdio
-      description: stdio, sse or streaming (default is stdio)
+      description: |
+        stdio, sse or streaming. Uses MCP_GATEWAY_AUTH_TOKEN environment variable for localhost authentication to prevent dns rebinding attacks.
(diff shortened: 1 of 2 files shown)


blah.json
saucow: @slimslenderslacks - did we want to include this, or is this a dev/test file?

PR #207, opened 2025-10-28
update the output of "docker mcp server ls" to surface more details

austin5456:
**What I did**
* Update the return form `docker mcp server ls` to also display OAuth, secret, config, and description

image

**Breaking Change**
This is breaking as it changes the data output of `docker mcp server ls` from an array of strings to an array of objects

**Related issue**
This is a precursor to some flow updates that smooth out setting up configurations after enabling a server

### DIFF
--- a/cmd/docker-mcp/commands/server.go
+++ b/cmd/docker-mcp/commands/server.go
@@ -209,11 +209,12 @@ func truncateString(s string, maxWidth int) string {
 	if maxWidth <= 0 {
 		return ""
 	}
-	if len(s) <= maxWidth {
+	runes := []rune(s)
+	if len(runes) <= maxWidth {
 		return s
 	}
 	if maxWidth > 3 {
-		return s[:maxWidth-3] + "..."
+		return string(runes[:maxWidth-3]) + "..."
 	}
-	return s[:maxWidth]
+	return string(runes[:maxWidth])
 }


cmd/docker-mcp/server/ls.go
```diff
@@ -18,5 +41,257 @@ func List(ctx context.Context, docker docker.Client) ([]string, error) {
...
+		return nil, err
+	}
+
+	// Get the list of configured secrets
+	configuredSecrets, err := desktop.NewSecretsClient().ListJfsSecrets(ctx)
+	if err != nil {
+		// If we can't get secrets, assume none are configured
+		if !quiet {
+			fmt.Fprintf(os.Stderr, "Warning: error fetching secrets: %v\n", err)
+		}
```
xenoscopic: Should we return `err` here?

saucow:
> I can't really speak to the correctness of the OAuth display logic (@saucow or @kgprs might have better input there), but a few things caught my eye:

@austin5456 - For OAuth DCR servers the logic is correct. The criteria is `remote` and if a `provider` is present. I tried out your branch locally and seeing the expected output

cmd/docker-mcp/server/ls.go
austin5456: Ah I think I see what you mean, secrets and OAuth just silently fail and log the error to the terminal ( if not in JSON format )
Not sure how hard of an error we want for them, specially OAuth if it fails to fetch for X or Y reasons

austin5456: @saucow Currently for getting the OAuth and Secrets we silently fail and print out a warning. A lot of the FE relies on the base return of the server names, since I'm not super well versed on how getting the OAuth and Secrets behave... is this overkill? should I instead just break the command? or are they likely to fail for external reasons?

PR #211, opened 2025-10-30
Add server search feature to workingsets

bobbyhouse:
**What I did**
Add ability to search across all working sets to return a list of servers grouped by working set id.

For example
`docker mcp workingset servers --filter test --workingset test-set`

Would return 
```
{
"id": 'test-set',
"servers":[{
    "type": "image"
    "image": "roberthouse224/test-set
  }]
}
```

Screenshot 2025-10-30 at 2 21 36 PM


cmrigney (approved):
lgtm. One minor nit, but not a blocker.

PR #213, opened 2025-10-31
Working sets: resolve server schemas/snapshots and cache them

cmrigney:
**What I did**
In preparation for getting working sets fully running in the gateway, this resolves the schema for running the server (as `catalog.Server` for now since the gateway knows how to run that).
- The resolved schema (aka snapshot) is cached in the DB with the server
- We include the snapshot when we `docker mcp workingset show my-set --format json/yaml` but remove it on export and pull (the thought was that the file should be more user-editable and including the snapshot would make it fragile)
- Server snapshots are resolved when a working set is either imported from a file, pulled, or first created.

A couple other notes:
- **Gap**: The registry snapshot is not implemented yet, just the self-describing image snapshot. I'll revisit this.
- **Breaking**: All images used must be self-describing. We can debate if we want to allow non-self-describing images (in the case when no config is needed). Thoughts @bobbyhouse ?

**Related issue**

### DIFF
--- a/pkg/workingset/import.go
+++ b/pkg/workingset/import.go
@@ -36,7 +36,7 @@ func Import(ctx context.Context, dao db.DAO, ociService oci.Service, filename st
 
 	// Resolve snapshots for each server before saving
 	for i := range len(workingSet.Servers) {
-		snapshot, err := resolveSnapshot(ctx, ociService, workingSet.Servers[i])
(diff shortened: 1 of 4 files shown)


bobbyhouse:
Nice work!

I agree on going with all images must be self-describing for workingsets. We may want to add a function to make creating that metadata easy in the future.

PR #222, opened 2025-11-06
Address CI Issues

saucow:
**What I did**
- This [commit](https://github.com/docker/mcp-gateway/commit/6b076b2479d8d1345c50c112119c62978d46858e) was merged from a private fork. The private fork was based on an old branch of `main` when we had logic in: `authenticationMiddleware` which was verifying the existence of: `DOCKER_MCP_IN_CONTAINER`. Since this logic is now in run.go [here](https://github.com/docker/mcp-gateway/blob/main/pkg/gateway/run.go#L334-L343) this test: `TestAuthenticationMiddleware_ContainerMode` is no longer relevant.
- Resolves lint issue on `main` branch

**Related issue**
- Fixes main CI issues

### DIFF
--- a/cmd/docker-mcp/catalog/reset_test.go
+++ b/cmd/docker-mcp/catalog/reset_test.go
@@ -1,7 +1,6 @@
 package catalog
 
 import (
-	"context"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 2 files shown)


cmrigney (approved):
Thanks for the quick fix!

PR #227, opened 2025-11-09
feat: List workingsets and clients

bobbyhouse:
**What I did**
Show which workingsets a client is connected to as part of it's JSON output.

I added additional output to the `docker mcp client ls -g --json` command that will additionally show which workingsets the client is connected to.

For example, if you connected Zed to the workingset named default you would have the following output for Zed:
```
  "zed": {
    "displayName": "Zed",
    "source": "https://zed.dev/",
    "icon": "https://raw.githubusercontent.com/docker/mcp-gateway/main/img/client/zed.png",
    "configName": "zed",
    "dockerMCPCatalogConnected": true,
    "workingsets": [
      "default"
    ],
    "error": null,
    "isInstalled": true,
    "isOsSupported": true
  }
}
```

**Related issue**

Screenshot 2025-11-09 at 2 01 41 PM

### DIFF
--- a/cmd/docker-mcp/client/parse.go
+++ b/cmd/docker-mcp/client/parse.go
@@ -38,7 +38,7 @@ func (c *MCPServerSTDIO) String() string {
 }
 
 func (c *MCPServerSTDIO) GetWorkingSet() string {
-	for i := 0; i < len(c.Args); i++ {
+	for i := range len(c.Args) {
 		arg := c.Args[i]
 		if arg == "--working-set" || arg == "-w" {
 			if i+1 < len(c.Args) {


cmd/docker-mcp/client/global_test.go
```diff
@@ -211,3 +211,57 @@ func TestGlobalCfgProcessor_SinglePath(t *testing.T) {
...
+	assert.True(t, result.IsMCPCatalogConnected)
+	require.Len(t, result.WorkingSets, 1)
+	assert.Equal(t, "test-set", result.WorkingSets[0])
+}
+
+func TestGlobalCfgProcessor_MultipleWorkingSets(t *testing.T) {
+	tempDir := t.TempDir()
+	configPath := filepath.Join(tempDir, "config.json")
+
+	require.NoError(t, os.WriteFile(configPath, []byte(`{"mcpServers": {"MCP_DOCKER": {"command": "docker", "args": ["mcp", "gateway", "run", "--working-set", "ws1", "--working-set", "ws2"]}}}`), 0o644))
```
cmrigney: **nitpick:** We don't support this case in the gateway and I'm not sure if we will, given the settings could conflict. Should we disallow this case?

cmrigney (approved):

PR #235, opened 2025-11-12
Mcp centralize config mapping logic

austin5456:
**What I did**
* ~Add prompts to add configs and secrets when enabling a server~
* Centralize the config and secrets logic ~now that two slices are using them~
* Fix a bug where the config values failed to update in certain situations

Reverting the prompt logic and commenting out the config functions in prep for working sets

**Related issue**

### DIFF
--- a/cmd/docker-mcp/server/registry_config.go
+++ b/cmd/docker-mcp/server/registry_config.go
@@ -9,14 +9,9 @@ import (
 	"github.com/docker/mcp-gateway/pkg/docker"
 )
 
-// loadRegistryWithConfig reads the registry and populates it with user config values.
-// It returns the populated registry and the userConfig map for further use.
 // IMPORTANT: config.yaml is the single source of truth for all config values.
 // We completely ignore any Config values that may exist in registry.yaml.
 func loadRegistryWithConfig(ctx context.Context, docker docker.Client) (config.Registry, map[string]map[string]any, error) {
-	// Read registry.yaml that contains which servers are enabled.
-	// NOTE: We only use this to determine which servers are enabled (the map keys).
-	// We completely ignore any Config values in the Tile structs from registry.yaml.
 	registryYAML, err := config.ReadRegistry(ctx, docker)
 	if err != nil {
 		return config.Registry{}, nil, fmt.Errorf("reading registry config: %w", err)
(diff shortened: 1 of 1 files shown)


cmd/docker-mcp/server/enable.go
```diff
@@ -114,3 +151,442 @@ func update(ctx context.Context, docker docker.Client, dockerCli command.Cli, ad
...
+		return field.Default, nil
+	}
+
+	// If input is empty and no default, return sentinel value to indicate skip
+	if input == "" {
+		return skipConfigValue, nil
+	}
+
+	// Validate input length to prevent extremely long values (DoS protection)
+	const maxInputLength = 10000 // Reasonable limit for config values
```
austin5456: is this... too long still?

cmd/docker-mcp/server/enable.go
```diff
@@ -114,3 +166,259 @@ func update(ctx context.Context, docker docker.Client, dockerCli command.Cli, ad
...
+			setNestedConfig(userConfig, field.Key, value)
+		}
+	}
+
+	fmt.Println()
+	return nil
+}
+
+// promptForConfigField prompts the user for a config field value
+func promptForConfigField(ctx context.Context, dockerCli command.Cli, field configField) (any, error) {
```
slimslenderslacks: I don't think this UX will work for all supported configs.  When enabling a server, we have to validate the current configuration against the json schema definition in the Tile.  This will complain if the current config is not valid (eg missing required fields).  And then I wish we could just put them in an editor with the json document (like git would do after a git commit), and then keep validating that config until it's good. Because what about arrays?  And incorrect formats?  It will be very hard to ensure that the schema will actually be valid after this flow.

docs/generator/reference/mcp_server_enable.md
```diff
@@ -7,6 +7,12 @@ Enable a server or multiple servers
 
 `docker mcp server enable`, `docker mcp server add`
 
+### Options
+
+| Name            | Type   | Default | Description                            |
+|:----------------|:-------|:--------|:---------------------------------------|
+| `--skip-config` | `bool` |         | Skip interactive configuration prompts |
```
slimslenderslacks: will Docker Desktop call `docker mcp server enable --skip-config`?

austin5456: yep https://github.com/docker/pinata/pull/37804/files

PR #259, opened 2025-11-21
Migrate legacy setup to default profile

cmrigney:
**What I did**

Implemented a migration path from the existing default `registry.yaml`, `config.yaml`, `tools.yaml` to a `default` profile.

Migrates when the profiles feature flag is on, only once ever (unless the db is deleted), either when:
- The gateway is ran with the `--profile` flag
- A `catalog-next` command is run
- A `profile` command is run

Once the migration is successful, it moves the legacy files to `~/.docker/mcp/.backup`.

**Follow up needed:** We'll need to migrate `remote` servers as well once we've implemented remote support in profiles.

**Related issue**

### DIFF
--- a/pkg/migrate/migrate.go
+++ b/pkg/migrate/migrate.go
@@ -233,7 +233,7 @@ func backupLegacyFiles() error {
 	_ = moveFile(catalogIndexPath, filepath.Join(backupDir, "catalog.json"))
 	_ = moveFile(oldCatalogPath, filepath.Join(backupDir, legacycatalog.DockerCatalogFilename))
 
-	// We use os.Remove to remove the directory, so only it's only removed if empty
+	// We use os.Remove to remove the directory, so it's only removed if empty
 	// We don't want to remove any custom catalog yamls the user may have added
 	_ = os.Remove(catalogsDir)
 


pkg/migrate/migrate.go
```diff
@@ -0,0 +1,228 @@
...
+	} else {
+		logs = append(logs, "no existing installed servers found, skipping default profile creation")
+	}
+
+	// Migration considered successful by this point
+	status = MigrationStatusSuccess
+
+	err = deleteLegacyFiles()
+	if err != nil {
+		logs = append(logs, fmt.Sprintf("failed to delete legacy files: %s", err.Error()))
```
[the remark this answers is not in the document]
cmrigney: Deleting the files is only best-effort, so this is intentional.

>  That makes recovery impossible without manually wiping the DB and rerunning the migration.

False conclusion?

pkg/migrate/migrate.go
```diff
@@ -0,0 +1,228 @@
...
+
+	registry, cfg, tools, oldCatalog, err := readLegacyDefaults(ctx, docker)
+	if err != nil {
+		logs = append(logs, fmt.Sprintf("failed to read legacy defaults: %s", err.Error()))
+		// Failed migration
+		return
+	}
+
+	// Only create a default profile if there are existing installed servers
+	if len(registry.ServerNames()) > 0 {
```
bobbyhouse: **Question**
I think I have been benefiting from a happy accident where I've been able to add/remove Atlassian for example, and then my config just pops back up and keeps working, but I suppose with profiles, since the config is scoped at the server we can't add the config to a server that doesn't exist?

bobbyhouse (approved):

pkg/migrate/migrate.go
cmrigney: Right. That will be one tradeoff with the profiles model. If that becomes a desired "feature" we could brainstorm some ideas.

cmrigney:
@bobbyhouse Good callout. That is [handled here](https://github.com/docker/mcp-gateway/pull/260#issue-3653132884).

> `docker mcp gateway` run will now default to attempting to run the profile with id `default`

bobbyhouse (approved):

PR #261, opened 2025-11-21
feat: Add remotes from legacy catalog

bobbyhouse:
**What I did**

Add support for remotes. For now, this only allows us to add the remotes that are in the catalog. You cannot add a remote directly by specifying a URL during profile creation or after the fact. The remote must exist in the catalog.

Add ability to represent the remote servers in our legacy catalog in the new catalog format as well as in working sets.

- Add new `Type` 'remote' to `WorkingSet.Server and `Catalog.Server`
- Add new property `endpoint` to `Catalog.Server`

**Related issue**

### DIFF
--- a/pkg/catalog_next/create_test.go
+++ b/pkg/catalog_next/create_test.go
@@ -717,11 +717,8 @@ func TestCreateFromLegacyCatalogWithRemotes(t *testing.T) {
 			require.NotNil(t, server.Snapshot)
 			assert.Equal(t, "test-server", server.Snapshot.Server.Name)
 
-			// Verify endpoint is set for remote servers
-			if server.Type == workingset.ServerTypeRemote {
-				assert.NotEmpty(t, server.Endpoint, "Endpoint should be set for remote servers")
-				assert.Equal(t, server.Snapshot.Server.Remote.URL, server.Endpoint, "Endpoint should match the Remote.URL from snapshot")
-			}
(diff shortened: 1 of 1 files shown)


bobbyhouse:
> ### 💡 Codex Review
> 
> https://github.com/docker/mcp-gateway/blob/6d1ce7a03fd3d253fe09a48c5c1bfc0363e608c8/pkg/catalog_next/create.go#L79-L86
> 
> **P1 Badge Preserve remote endpoints when exporting working sets**
> 
> When building a catalog from an existing working set, the loop creating `servers` copies type, tools, source, image, and snapshot but never copies `server.Endpoint`. If the working set contains a remote server, the resulting catalog entry has an empty endpoint even though the struct validator (see `Catalog.Server` with `required_if=Type remote`) requires it, so `Create` fails validation for any working set that includes a remote server. Please propagate the endpoint from the working set into the catalog server here.
> ℹ️ About Codex in GitHub
> 
> [Your team has set up Codex to review pull requests in this repo](http://chatgpt.com/codex/settings/general). Reviews are triggered when you
> 
>     * Open a pull request for review
> 
>     * Mark a draft as ready
> 
>     * Comment "@codex review".
> 
> 
> If Codex has suggestions, it will comment; otherwise it will react with 👍.
> 
> Codex can also answer questions or update the PR. Try commenting "@codex address that feedback".

ah, yeah, nice catch. Pushing up a fix.

Daniel-Kolev (approved):
Cody's changes lgtm

PR #267, opened 2025-11-25
Add Cline to MCP client configurations

saoudrizwan:
**What I did**

Adds Cline (VS Code/Cursor extension) as a supported MCP client. Cline stores its MCP settings in the VS Code/Cursor globalStorage directory at `saoudrizwan.claude-dev/settings/cline_mcp_settings.json` and uses the standard `mcpServers` configuration format.

### DIFF
--- a/pkg/client/config.yml
+++ b/pkg/client/config.yml
@@ -300,3 +300,4 @@ project:
       list: '.mcpServers | to_entries | map(.value + {"name": .key})'
       set: .mcpServers[$NAME] = $JSON
       del: del(.mcpServers[$NAME])
+


pkg/client/config.yml
```diff
@@ -39,6 +39,30 @@ system:
       list: '.mcpServers | to_entries | map(.value + {"name": .key})'
       set: .mcpServers[$NAME] = $JSON
       del: del(.mcpServers[$NAME])
+  cline:
+    displayName: Cline
+    source: https://github.com/cline/cline
```
slimslenderslacks: This is probably fine but we show this url in the UI when we can not detect a cline install.  So if you want to use a url that jumps straight to install instructions, that could also be a good choice.

PR #270, opened 2025-12-02
Add --no-tools to catalog show and server ls 

bobbyhouse:
**What I did**

- Add command `docker mcp catalog-next server ls` with filtering options on the server name so that we can get the details of a single server, including all of it's tools
- Add a --no-tools option to `docker mcp catalog-next show` because we were returning too much data that would exceed a 1MB limit for node processes to for data over stdio
- Make `docker mcp catalog-next ls --format json/yaml` return list data instead of all the data for all the catalogs, which would also exceed the 1MB limit

**Related issue**

### DIFF
--- a/pkg/catalog_next/catalog.go
+++ b/pkg/catalog_next/catalog.go
@@ -25,6 +25,12 @@ type CatalogWithDigest struct {
 	Digest  string `yaml:"digest" json:"digest"`
 }
 
+type CatalogSummary struct {
+	Ref    string `yaml:"ref" json:"ref"`
+	Digest string `yaml:"digest" json:"digest"`
+	Title  string `yaml:"title" json:"title"`
+}
+
 // Source prefixes must be of the form "<prefix>:"
 const (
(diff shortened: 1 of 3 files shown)


Daniel-Kolev (approved):

pkg/catalog_next/show.go
```diff
@@ -77,6 +77,10 @@ func Show(ctx context.Context, dao db.DAO, ociService oci.Service, refStr string
 
 	catalog := NewFromDb(dbCatalog)
 
+	if noTools {
+		catalog = filterCatalogTools(catalog)
+	}
```
cmrigney: **suggestion (non-blocking):** I would personally find it more readable if we edit in-place to nil the values out instead of making a copy:

```suggestion
	if noTools {
		for i := range catalog.Servers {
			catalog.Servers[i].Tools = nil
			if catalog.Servers[i].Snapshot != nil {
				catalog.Servers[i].Snapshot.Server.Tools = nil
			}
		}
	}
```

PR #277, opened 2025-12-03
Disallow session flag with profile flag.

cmrigney:
**What I did**

Codex caught a conflict here: https://github.com/docker/mcp-gateway/pull/276#discussion_r2586180994

This just disables using `--session` with `--profile`.

**Related issue**


slimslenderslacks (changes requested):
I'm going to remove the session flag all together. I'll open a new PR.

PR #294, opened 2025-12-10
Fix migration errors caused by multiple simultaneous processes

cmrigney:
**What I did**

There were two race conditions I was seeing, both were leading to the same result. If two instances of the cli ran at the same time, they would both attempt migrations and one of them would fail. This wouldn't hurt the state of the database, but it would lead to failure output on one of the processes. The two cases were:
- Migrating the database schema
- Migrating the prior configuration to profiles

This PR adds retries for both and transactional design for the latter. So there should be no more errors caused by the race condition.

**Related issue**


cmrigney:
@codex

bobbyhouse (approved):

PR #298, opened 2025-12-10
Fix: Auto-detect Docker CE mode on Linux

### LINKED ISSUE #293: Failed to connect to OAuth notifications error in Docker CE
# Issue

It appears that you _must_ set the environment variable `DOCKER_MCP_USE_CE=true` so that the function [IsCEMode](https://github.com/docker/mcp-gateway/blob/main/pkg/oauth/mode.go#L9-L11) will return true. 

We should be able to use `docker.RunningInDockerCE` like is done [elsewhere](https://github.com/docker/mcp-gateway/blob/main/cmd/docker-mcp/commands/root.go#L71) in the project.

That way people running the mcp module in Docker CE will not have to manually set the env var.

# Instructions to reproduce

The following instructions will allow you to reproduce the issue.

I am running Debian 11 and followed instructions from https://docs.docker.com/engine/install/debian/ to install Docker CE (Cloud Engine). 

I pulled down the mcp-gateway project and built it (`make docker mcp`).

I run `docker mcp gateway run` and then get errors related to OAuth
```
Failed to connect to OAuth notifications: Get "http://localhost/notify/notifications/channel/external-oauth": dial unix /root/.docker/desktop/backend.sock: no such file or directory.
```

AryanBagade:
**What I did**
Auto-detect Docker CE mode on Linux without requiring users to manually set the `DOCKER_MCP_USE_CE=true` environment variable.

**Related issue**
Fixes #293

**Problem**
When running `docker mcp gateway run` on Linux with Docker CE, users encountered: Failed to connect to OAuth notifications: 
`Failed to connect to OAuth notifications: Get "http://localhost/notify/notifications/channel/external-oauth": dial unix /root/.docker/desktop/backend.sock: no such file or directory.`

This happened because [`IsCEMode()`](https://github.com/docker/mcp-gateway/blob/main/pkg/oauth/mode.go#L9-L11) only checked for the `DOCKER_MCP_USE_CE` environment variable. When not set, it returned `false`, causing the OAuth
notification monitor to attempt connecting to Docker Desktop's backend socket - which doesn't exist on Docker CE installations.

**Solution**
Modified `IsCEMode()` to auto-detect Docker CE on Linux by checking if Docker Desktop is running, following the same
pattern used in [`pkg/features/features.go#L63-L84`](https://github.com/docker/mcp-gateway/blob/main/pkg/features/features.go#L63-L84).

image

### DIFF
--- a/pkg/oauth/credhelper_test.go
+++ b/pkg/oauth/credhelper_test.go
@@ -10,50 +10,43 @@ import (
 )
 
 func TestIsCEMode(t *testing.T) {
-	// Test CE mode detection
-	tests := []struct {
-		name     string
-		envValue string
-		expected bool
-	}{
-		{
-			name:     "CE mode enabled",
-			envValue: "true",
-			expected: true,
-		},
-		{
-			name:     "CE mode disabled",
-			envValue: "false",
-			expected: false,
-		},
-		{
-			name:     "CE mode not set",
-			envValue: "",
-			expected: false,
-		},
-	}
-
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			oldValue := os.Getenv("DOCKER_MCP_USE_CE")
-			defer func() {
-				if oldValue == "" {
-					os.Unsetenv("DOCKER_MCP_USE_CE")
-				} else {
-					os.Setenv("DOCKER_MCP_USE_CE", oldValue)
-				}
-			}()
-
-			if tt.envValue == "" {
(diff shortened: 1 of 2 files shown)


saucow:
@slimslenderslacks yes: https://github.com/docker/mcp-gateway/pull/301 is the same fix by exporting the existing: `IsRunningInDockerDesktop` with the addition of removing the reconnect as well. @AryanBagade thank you for bringing this up closing this

PR #302, opened 2025-12-11
Fix dynamic tools with profiles

cmrigney:
**What I did**

This loads the default `mcp/docker-mcp-catalog` catalog when using profiles so that dynamic tools can be used with it.

A couple caveats to this:
1. It requires pulling the catalog. This will happen automatically by DD. But non-DD users will have to manually pull it.
2. Dynamic tools can't be turned off for profiles except by the feature flag.

**Related issue**

### DIFF
--- a/pkg/gateway/configuration_workingset.go
+++ b/pkg/gateway/configuration_workingset.go
@@ -19,18 +19,16 @@ import (
 )
 
 type WorkingSetConfiguration struct {
-	WorkingSet   string
-	dynamicTools bool
-	ociService   oci.Service
-	docker       docker.Client
+	config     Config
+	ociService oci.Service
+	docker     docker.Client
 }
 
-func NewWorkingSetConfiguration(workingSet string, dynamicTools bool, ociService oci.Service, docker docker.Client) *WorkingSetConfiguration {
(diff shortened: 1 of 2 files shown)


pkg/gateway/configuration_workingset.go
```diff
@@ -101,10 +123,6 @@ func (c *WorkingSetConfiguration) readOnce(ctx context.Context, dao db.DAO) (Con
 
 		serverName := server.Snapshot.Server.Name
 
-		if _, exists := servers[serverName]; exists {
```
cmrigney: This was an erroneous check anyways. We already do this validation in a profile when it's saved.

bobbyhouse:
functionally tested this and it worked for me. Should review with @slimslenderslacks but I was able to use `mcp-find` to find a server.

pkg/gateway/configuration_workingset.go
```diff
@@ -18,16 +18,18 @@ import (
 )
 
 type WorkingSetConfiguration struct {
-	WorkingSet string
-	ociService oci.Service
-	docker     docker.Client
+	WorkingSet   string
```
slimslenderslacks: See below, you may just want the WorkingSetConfiguration to have the configuration.  We're certainly not done tweaking this.

slimslenderslacks (approved):
this is exactly what I was expecting here so looks good.

PR #313, opened 2025-12-23
Slim/dynamic profiles

slimslenderslacks:
**What I did**

**Related issue**

### DIFF
--- a/pkg/gateway/run.go
+++ b/pkg/gateway/run.go
@@ -253,7 +253,7 @@ func (g *Gateway) Run(ctx context.Context) error {
 			_, _ = req.Session.ListRoots(ctx, &mcp.ListRootsParams{})
 		},
 		CompletionHandler: nil,
-		InitializedHandler: func(ctx context.Context, req *mcp.InitializedRequest) {
+		InitializedHandler: func(_ context.Context, req *mcp.InitializedRequest) {
 			clientInfo := req.Session.InitializeParams().ClientInfo
 			log.Log(fmt.Sprintf("- Client initialized %s@%s %s", clientInfo.Name, clientInfo.Version, clientInfo.Title))
 
@@ -751,4 +751,3 @@ func (g *Gateway) profileLoadingMiddleware() mcp.Middleware {
 		}
 	}
 }
-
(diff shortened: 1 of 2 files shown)


pkg/gateway/run.go
```diff
@@ -251,9 +253,28 @@ func (g *Gateway) Run(ctx context.Context) error {
...
+				log.Log(fmt.Sprintf("- Current working directory: %s", pwd))
+			}
+
+			// Log entire initialize request
+			initParams := req.Session.InitializeParams()
+			if initParams != nil {
+				if initJSON, err := json.MarshalIndent(initParams, "  ", "  "); err == nil {
+					log.Log(fmt.Sprintf("- Initialize request:\n  %s", string(initJSON)))
+				}
+			}
```
cmrigney: **question:** Is this just debugging code? Should we remove it so we don't bloat the logs?

.claude/settings.json
cmrigney: **question:** Just double checking, is committing this intentional?

pkg/gateway/activateprofile.go
```diff
@@ -0,0 +1,321 @@
...
+		// Temporarily add server to configuration to fetch updated secrets
+		originalServerNames := slices.Clone(g.configuration.serverNames)
+		g.configuration.serverNames = append(g.configuration.serverNames, serverName)
+
+		// Add server to servers map for secret resolution
+		g.configuration.servers[serverName] = serverConfig
+
+		// Fetch updated secrets for validation
+		if g.configurator != nil {
+			updatedSecrets, err := g.configurator.readDockerDesktopSecrets(ctx, g.configuration.servers, g.configuration.serverNames)
```
cmrigney: **nitpick:** Should we log the error here?

cmrigney (commented):
I would be interested in seeing a demo of these changes!

pkg/gateway/run.go
slimslenderslacks: I think we need this for now.

pkg/gateway/activateprofile.go
slimslenderslacks: sounds good

.claude/settings.json
slimslenderslacks: yes

pkg/gateway/activateprofile.go
```diff
@@ -0,0 +1,326 @@
...
+// ActivateProfile activates a profile by name, loading its servers into the gateway
+func (g *Gateway) ActivateProfile(ctx context.Context, profileName string) error {
+	// Load profile from database
+	dao, err := db.New()
+	if dao != nil {
+		defer dao.Close()
+	}
+	if err != nil {
+		return fmt.Errorf("failed to create database client: %w", err)
+	}
```
cmrigney: **nitpick (nonblocking):** fwiw, a more idiomatic pattern would be:

```suggestion
	if err != nil {
		return fmt.Errorf("failed to create database client: %w", err)
	}
	defer dao.Close()
```

PR #323, opened 2026-01-06
Add support for extraHosts in MCP server catalog entries

slimslenderslacks:
Add support for the extraHosts setting in Catalog entries which allows specifying custom hostname-to-IP mappings for MCP containers. When present, the gateway passes these entries to Docker using the --add-host flag, adding static entries to the container's /etc/hosts file.

Changes:
- Add ExtraHosts []string field to catalog.Server struct
- Add --add-host flag support in clientpool.argsAndEnv() function
- Add TestApplyConfigExtraHosts test with playwright-like catalog entry
- Add example catalog file showing usage with playwright server

Example usage in catalog YAML:
  playwright: image: mcp/playwright@sha256:... extraHosts: - "myapp.local:192.168.1.100" - "testserver:10.0.0.5"

### DIFF
--- a/examples/playwright-with-extrahosts.yaml
+++ b/examples/playwright-with-extrahosts.yaml
@@ -1,0 +1,17 @@
+registry:
+  playwright:
+    description: Playwright MCP server.
+    title: Playwright
+    type: server
+    longLived: true
+    dateAdded: "2025-05-05T20:04:34Z"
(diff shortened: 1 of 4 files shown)


cmrigney (approved):
lgtm. Made a non-blocking suggestion about moving the example.

PR #332, opened 2026-01-08
Add missing fields to server catalog data

cmrigney:
**What I did**

New catalogs and profiles are missing several fields that are provided in our hosted legacy catalog.
1. The readme url
2. Additional tool info. This originally comes from a tools url, but in the profiles world we'll package it in with the server snapshot.

This is prep work that will enable us to fix a bug we have in Docker Desktop where `server inspect` inspects the old catalog instead of the new.

**Related issue**


cmrigney:
@jchangx The legacy hosted catalog is here (still actively used in production): http://desktop.docker.com/mcp/catalog/v3/catalog.yaml

And there's also a loose spec of it [here](https://github.com/docker/mcp-gateway/blob/main/docs/server-entry-spec.md).  But this spec is going to drift from what's in the legacy catalog as we press forward toward profiles and oci catalogs. (Actually you've reminded me that I forgot to update it!)

PR #340, opened 2026-01-14
feat: Add telemetry for profiles and catalogs

bobbyhouse:
**What I did**
- Capture profile operations (e.g. create, push, pull)
- Capture catalog-next operations (e.g. create, push pull)
- Capture `--profile` arg passed to `docker-mcp gateway run`
- Added a skill to add this type of telemetry

**Related issue**

### DIFF
--- a/docs/telemetry/README.md
+++ b/docs/telemetry/README.md
@@ -108,12 +108,11 @@ Operations for managing MCP server configurations:
 - **`mcp.catalog.operation.duration`** - Duration of catalog operations
 - **`mcp.catalog.servers`** - Gauge showing number of servers in catalogs
 
-### Working Set Operations
+### Profile Operations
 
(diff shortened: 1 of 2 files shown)


pkg/telemetry/telemetry.go
bobbyhouse: Should we be consistent and stick with working set instead of profiles?

.claude/skills/otel-instrument/SKILL.md
cmrigney: **praise:** Awesome addition!

cmrigney (dismissed):
Lgtm! Looks like the docs also need updated. Feel free to ignore the nitpicks especially since the prior test code followed the same pattern.

cmrigney (approved):

PR #354, opened 2026-01-21
Changes to use Docker Desktop proxy

karman-docker:
**What I did**
Changed the default http client transport to use Docker Desktop proxy.

- Use Sync.Once to initialize `desktopProxyTransportInst` which uses http.DefaultTransport if Docker Desktop is not running, otherwise creates a transport that forwards the http requests via Docker Desktop proxy.
- I had to move `features.IsRunningInDockerDesktop()` from `features` package to `desktop` package. Because I cannot access `features` package in `desktop` package as it results in a circular dependency between those two packages.
- 

Fixes https://docker.atlassian.net/browse/DDB-378

**Related issue**
https://docker.atlassian.net/browse/DDB-378


karman-docker:
> Thanks for doing this! I think there are a couple things we need to handle first:
> 
>     1. There are other places where we construct an `http.Client` where we'd also want to use this proxy. If you search them in the codebase you should find them.
> 
>     2. We need a fallback for Docker CE and container contexts, since DD won't be available there. The solution is probably to make `DesktopProxyTransport` check `features.IsRunningInDockerDesktop()` and return the default transport if it's not running in DD, instead. All of this can be wrapped in a `sync.Once` so we don't redundantly check DD presence (e.g. check out `pkg/validate/validate.go` for an example).

I referred to your PR https://github.com/docker/mcp-gateway/pull/118 to patch the http.Client usages. Yes there are few other source files that use http.Client
```
./pkg/oauth/notification_monitor.go
./pkg/catalog/catalog.go
./pkg/oci/import.go
./pkg/desktop/raw_client.go
./pkg/mcp/remote.go
./pkg/fetch/fetch.go
./pkg/registryapi/client.go
./pkg/integration_test.go
./pkg/gateway/findtools.go
./pkg/gateway/registry.go
./pkg/gateway/mcpadd.go
./pkg/gateway/auth_integration_test.go
```

I will update these files as well and use sync.Once(). 

I do not need to change the http.Client references in `vendor` folders correct ?

PR #359, opened 2026-01-23
fix(secret): fail fast when Docker Desktop is not available

### LINKED ISSUE #355: Set MCP secret in Docker engine
Greetings,

I installed the MCP toolkit on a headless machine running Docker engine (so no Docker Desktop) as explained [here](https://forums.docker.com/t/mcp-toolkit-in-docker-engine-on-linux/150888). I added the google-maps-comprehensive server, and I now need to add an API key for that. However, I failed so far:

```
$ docker mcp secret set GOOGLE_MAPS_API_KEY=my_api_key
Post "http://localhost/secrets": dial unix /root/.docker/desktop/jfs.sock: connect: no such file or directory
```

Note that, to avoid typing `sudo docker` every time, I aliased `docker` to `sudo docker`, but I don’t think this matters here. I know you can give permission to your user to run it without sudo, but this way seems to work just as well without requiring a relogin. It seems to me that the issue here is that the MCP toolkit is assuming a Docker Desktop installation.

Any idea how I can fix or work around this?
Thank you for your help,

GTP

Edit: this seems to be related to issue #119 , where another user shows an example of a configuration file to work around the problem. But it's not clear to me where I should place that file, and if it needs to have a specific name.

ghost:
## What I did

Introduced explicit environment validation and hardened secret handling in the `secret` package to ensure `docker mcp secret set` behaves correctly depending on the runtime environment.

This change clarifies the distinction between Docker Desktop and headless Docker Engine environments, failing fast when Desktop-only secret backends are unavailable. It also improves secret parsing, provider validation, and secure STDIN-based value ingestion, reducing ambiguity and preventing partially executed flows.

The implementation wires environment detection through the Docker CLI initialization path and centralizes provider semantics, ensuring clearer errors and safer secret management.

## Related issue

Fixed #355 

## What was the problem ? 

The `docker mcp secret set` command could be executed in environments where Docker Desktop services are not available (e.g. Docker Engine / CE).

In these cases:

- The command initialized successfully
- Execution proceeded into Desktop-dependent logic
- Failures occurred late in the flow
- Error messages did not clearly explain that Docker Desktop was required

This resulted in confusing behavior where the command appeared broken instead of explicitly unsupported.

### Secret Set Execution & Environment Validation Flow

```mermaid
flowchart TD
    A[User runs docker mcp secret set] --> B[Parse provider & secret input]
    B --> C[Initialize Docker CLI]
    C --> D{Running on Docker Desktop?}
    D -->|Yes| E[Set secret via Desktop / JFS backend]
    D -->|No| F[Fail fast with clear error]
```

## Explanation 

The diagram illustrates the new execution flow for `docker mcp secret set`:

#### Input Handling
The command first parses secret input, supporting both `key=value` syntax and secure STDIN-based ingestion depending on the selected provider.

#### Environment Detection
After initializing the Docker CLI, the runtime environment is explicitly checked to determine whether Docker Desktop services are available.

#### Fail-Fast Behavior
If the command is executed on Docker Engine (headless / CE), execution stops immediately with a clear, actionable error message.

#### Desktop Path
When Docker Desktop is available, the secret is safely stored using the Desktop-backed secrets client.

This ensures unsupported environments are rejected early, while supported environments follow the normal happy path.

## Before

- No explicit validation of Docker Desktop availability
- Command executed partially in unsupported environments
- Errors surfaced late and without clear guidance
- Users had no indication that Docker Desktop was required

## After

- Runtime environment is validated early
- Headless Docker Engine is detected explicitly
- Unsupported environments fail fast with actionable errors
- Desktop environments continue to work unchanged
- Secret parsing and provider semantics are clearer and safer

## Testing

This change was validated through local builds and code review.
No new automated tests were added as part of this PR, since the behavior depends on runtime environment detection (Docker Desktop vs. headless Docker Engine) and would require environment-specific integration coverage.
Follow-up work may introduce targeted integration tests once a suitable test harness for Docker Desktop–dependent behavior is available.

### DIFF
--- a/cmd/docker-mcp/secret-management/secret/set.go
+++ b/cmd/docker-mcp/secret-management/secret/set.go
@@ -114,8 +114,8 @@ func Set(ctx context.Context, s Secret, opts SetOpts) error {
 
 	if isCE {
 		return fmt.Errorf(
-			"Docker Desktop is not available. " +
-				"`docker mcp secret set` requires Docker Desktop to manage secrets. " +
+			"Docker Desktop is not available " +
+				"`docker mcp secret set` requires Docker Desktop to manage secrets " +
 				"If you are running Docker Engine in a headless environment, " +
 				"use --secrets with a .env file instead",
 		)
--- a/cmd/docker-mcp/secret-management/secret/set.go
+++ b/cmd/docker-mcp/secret-management/secret/set.go
@@ -115,9 +115,9 @@ func Set(ctx context.Context, s Secret, opts SetOpts) error {
 
 	if isCE {
 		return fmt.Errorf(
-			"Docker Desktop is not available " +
-				"`docker mcp secret set` requires Docker Desktop to manage secrets " +
-				"If you are running Docker Engine in a headless environment, " +
+			"docker desktop is not available: " +
+				"`docker mcp secret set` requires docker desktop to manage secrets; " +
+				"if you are running docker engine in a headless environment, " +
 				"use --secrets with a .env file instead",
 		)
 	}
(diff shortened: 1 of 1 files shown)


cutecatfann (commented):
LGTM!

cutecatfann (dismissed):

cutecatfann (dismissed):

cutecatfann (dismissed):

cutecatfann (approved):

PR #361, opened 2026-01-23
client: mark Amazon Q as non-stdio MCP client

### LINKED ISSUE #353: Add support for Amazon Q Developer client
Amazon Q is not in the supported clients list. Adding support for Amazon Q would enable Docker MCP integration for AWS-focused development workflows.

ghost:
## What I did

Added Amazon Q Developer to the list of supported MCP clients in the Docker MCP Gateway configuration.
This change registers Amazon Q as a non-stdio MCP client, making it visible, discoverable, and correctly handled by the gateway while explicitly blocking unsupported stdio connection attempts.
The goal is to enable Docker MCP integration visibility for AWS-focused workflows without allowing invalid transport usage.

## Related issue

Fixed #353

## What was the problem ?

Amazon Q Developer was not present in the supported clients list.

Because of this:

- Amazon Q did not appear in docker-mcp client ls
- Users had no official confirmation that Amazon Q was recognized by the gateway
- There was no explicit signal that Amazon Q does not support MCP stdio
- Invalid connection attempts failed implicitly instead of by design

## Explanation

Amazon Q Developer does not support MCP stdio transports.

This PR explicitly registers Amazon Q with:

- stdio: false
- Proper client metadata
- Gateway-level enforcement of transport limitations

By doing so, the gateway can expose Amazon Q safely while preventing unsupported usage paths.

### Amazon Q Client Registration & Connection Flow

```mermaid
flowchart TD
    A[MCP Gateway starts] --> B[Load supported clients]
    B --> C{Client is Amazon Q?}
    C -->|Yes| D[Register client with stdio disabled]
    D --> E[Expose client via client ls]
    E --> F{User attempts connect?}
    F -->|Yes| G[Reject: stdio not supported]
    F -->|No| H[Client remains discoverable]
```

## Before

- Amazon Q was not listed as a supported client
- docker-mcp client ls did not show Amazon Q
- No transport constraints were enforced at registration level
- Users could not validate Amazon Q MCP support

## After

- Amazon Q appears in docker-mcp client ls
- Client is marked as disconnected by default
- Stdio connections are explicitly rejected with a clear error
- Gateway behavior matches Amazon Q’s real MCP capabilities

## Configuration (config.yml)

Amazon Q is explicitly declared in the embedded config.yml file.
The configuration registers the client with full metadata while intentionally disabling stdio support, making the limitation explicit and enforceable at the configuration level.

image

### Client discovery

```go run ./cmd/docker-mcp client ls --global```

Observed behavior:

- amazon-q appears in the system-wide MCP configurations
- The client is correctly marked as disconnected
- Other supported clients (for example, VS Code) remain unaffected
- This confirms that Amazon Q is correctly registered and discoverable by the MCP Gateway.

### Connection attempt

```go run ./cmd/docker-mcp client connect amazon-q --global```

Observed behavior:

```client does not support MCP stdio```

This shows that:

- The gateway explicitly blocks stdio connections for Amazon Q
- The failure is intentional and enforced by configuration
- Unsupported transport usage is rejected with a clear error message

### State validation after failure

After the failed connection attempt, the client list was queried again:

```go run ./cmd/docker-mcp client ls --global```

Observed behavior:

- amazon-q remains disconnected
- No partial or inconsistent state is introduced
- Other clients remain connected and functional

This demonstrates that invalid connection attempts do not corrupt client state and that transport enforcement behaves as expected.

image

### DIFF
--- a/docs/generator/reference/docker_mcp_client_connect.yaml
+++ b/docs/generator/reference/docker_mcp_client_connect.yaml
@@ -1,12 +1,12 @@
 command: docker mcp client connect
 short: |
-    Connect the Docker MCP Toolkit to a client. Supported clients: claude-code claude-desktop cline codex continue cursor gemini goose gordon kiro lmstudio opencode sema4 vscode zed
+    Connect the Docker MCP Toolkit to a client. Supported clients: amazon-q claude-code claude-desktop cline codex continue cursor gemini goose gordon kiro lmstudio opencode sema4 vscode zed
 long: |
-    Connect the Docker MCP Toolkit to a client. Supported clients: claude-code claude-desktop cline codex continue cursor gemini goose gordon kiro lmstudio opencode sema4 vscode zed
+    Connect the Docker MCP Toolkit to a client. Supported clients: amazon-q claude-code claude-desktop cline codex continue cursor gemini goose gordon kiro lmstudio opencode sema4 vscode zed
 usage: |-
     docker mcp client connect [OPTIONS] <mcp-client>
 
-    Supported clients: claude-code claude-desktop cline codex continue cursor gemini goose gordon kiro lmstudio opencode sema4 vscode zed
+    Supported clients: amazon-q claude-code claude-desktop cline codex continue cursor gemini goose gordon kiro lmstudio opencode sema4 vscode zed
 pname: docker mcp client
 plink: docker_mcp_client.yaml
 options:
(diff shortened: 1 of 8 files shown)


cutecatfann (dismissed):
Thanks for fixing! 😁

cutecatfann (approved):
Thanks for patience! Looks great to me!

PR #364, opened 2026-01-23
fix: avoid false positives when blocking secrets in tool responses

### LINKED ISSUE #334: Secret blocking false positives
The Linear MCP server through Docker MCP Toolkit is consistently reporting that secrets are being returned when using tools like "get_issue" (which reads the issue).

It seems like the secret blocking heuristics are a bit too restrictive in this case. I've found the code block that seems to be generating the message I'm seeing:

https://github.com/docker/mcp-gateway/blob/83c8ab557ee6fb23bd2294f5c3a70dee07aa7ca7/pkg/interceptors/block_secrets.go#L14

Are there any workarounds?

ghost:
## What I did

Adjusted the secret-blocking middleware to avoid false positives for read-only tools by scoping response secret scanning to cases where secrets were already present in the tool input.
This change ensures that tools such as `get_issue` no longer fail due to overly aggressive secret detection in responses, while preserving protection against real secret exfiltration scenarios.

## Related issue

Fixed #334 

## What was the problem?

When using read-only tools (e.g. `get_issue`) through the MCP Gateway, users consistently hit errors indicating that secrets were being returned, even though no secrets were provided as input.

This behavior was caused by the secret-blocking middleware scanning **all tool responses unconditionally**, regardless of whether the request itself contained sensitive data.

As a result:

- Read-only tools were blocked unexpectedly
- Tools that simply return text (issues, tickets, descriptions) failed
- Users had no reliable workaround other than disabling secret blocking entirely

## How this change fixes it
The middleware now tracks whether secrets were present in the tool input and only scans the tool response when there is a real risk of secret exfiltration.

Specifically:

- Tool arguments are scanned first
- If secrets are detected in the input, the request is rejected as before
- Tool responses are only scanned **if secrets were present in the input**
- Read-only tools with non-sensitive inputs are no longer blocked

This preserves the original security guarantees while eliminating false positives.

### Before
- Tool responses were always scanned for secrets
- Read-only tools like `get_issue` could fail unexpectedly
- Frequent false positives from heuristic secret detection

### After
- Tool responses are scanned only when secrets were present in the input
- Read-only tools work reliably
- Protection against secret exfiltration is preserved

### Secret detection in tool call input
This section ensures that tool invocation arguments are scanned for secrets before the tool is executed.
The arguments are first normalized into a string representation, allowing them to be safely analyzed regardless of their original structure. If the resulting input is non-empty and a secret pattern is detected, the execution is immediately blocked.

image

### Conditional secret scanning in tool responses
This logic ensures that tool responses are only scanned for secrets when the input originally contained sensitive data.
Instead of blindly inspecting every tool response, the middleware tracks whether a secret was detected in the input arguments (inputHadSecrets). Only in that case does it perform a secondary scan on the tool’s output.

image

### Tracking secret presence across the tool call lifecycle
The inputHadSecrets flag is introduced to track whether sensitive data was detected in the tool input arguments.
This variable acts as state shared across the middleware execution, allowing later stages to make informed decisions based on what happened during input validation.

image

### DIFF
--- a/pkg/interceptors/block_secrets.go
+++ b/pkg/interceptors/block_secrets.go
@@ -21,6 +21,7 @@ func BlockSecretsMiddleware() mcp.Middleware {
 
 			var toolName string
 			var arguments any
+			var inputHadSecrets bool
 
 			// Try to extract from request
 			if callReq, ok := req.(*mcp.CallToolRequest); ok && callReq.Params != nil {
@@ -32,7 +33,8 @@ func BlockSecretsMiddleware() mcp.Middleware {
 				log.Logf("  - Scanning tool call arguments for secrets...\n")
 
 				argumentsStr := argumentsToString(arguments)
-				if secretsscan.ContainsSecrets(argumentsStr) {
+				if argumentsStr != "" && secretsscan.ContainsSecrets(argumentsStr) {
+					inputHadSecrets = true
 					return nil, fmt.Errorf("a secret is being passed to tool %s", toolName)
 				}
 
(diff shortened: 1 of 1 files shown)


cutecatfann (approved):
Thank you!

PR #365, opened 2026-01-24
Parse JSON arguments correctly in docker mcp tools call

### LINKED ISSUE #307: docker mcp tools call Fails to Parse JSON Arguments, Treating Complex Objects as Strings
### **Summary**

The `docker mcp tools call` command line utility fails to correctly parse arguments provided as complex JSON objects (e.g., `'{"key": "value"}'`). Instead of deserializing the string into a nested object/map, the Go CLI passes the literal JSON string to the tool handler. This leads to validation errors in downstream APIs, as the API wrapper attempts to process the JSON string's *characters* as if they were property keys.

### **Affected Component(s)**

* `mcp-cli` / `docker mcp tools call`
* File: `.../cli/tools/call.go` (specifically the `parseArgs` function)

### **Version**

* MCP Gateway/CLI Version: **(Please Specify Version)**
* Go runtime version: **(Please Specify Version)**

### **Steps to Reproduce**

1.  Set up a tool (e.g., `API-patch-page`) that expects a complex JSON object for an argument (e.g., `properties`).
2.  Attempt to call the tool with a valid, quoted JSON payload.

    ```bash
    docker mcp tools call API-patch-page \
    page_id='<VALID_UUID>' \
    properties='{"Test":{"rich_text":[{"text":{"content":"test content"}}]}}'
    ```

3.  Alternatively, use the simplest JSON object:

    ```bash
    docker mcp tools call API-patch-page page_id='<VALID_UUID>' properties='{}'
    ```

### **Observed Behavior**

The command fails with a `validation_error` because the tool receives the `properties` value as a literal string (`"{}"`) and tries to process its characters (`{` and `}`) as property identifiers.

* **Error Message (for `properties='{}'`):**
    ```json
    {"status":400,"object":"error","code":"validation_error","message":"Couldn't find editable properties for these identifiers: 0,1","request_id":"..."}
    ```
* **Root Cause:** The identifiers `0` and `1` correspond to the character indices of the string `"{"}"` which the tool is incorrectly trying to use as property names.

### **Technical Analysis (Root Cause)**

The issue is that the `parseArgs` function only assigns the value as a raw string without checking if it contains JSON.

The problematic code in `parseArgs`:

```go
// ... inside parseArgs function ...

if len(parts) == 2 {
    key = parts[0]
    value = parts[1] // <-- value is always a Go string here, even if it contains JSON.
} 

// ...

```
### **Proposed Solution (Fix in `call.go`)**

The `parseArgs` function must attempt to unmarshal (decode) the string value as JSON. If the unmarshalling is successful, the resulting complex object should be used; otherwise, the value should remain a simple string. 

The corrected logic for the value assignment block in `parseArgs` should look like this:

```go
import (
    "encoding/json" 
    "strings"
)

// ... inside parseArgs function ...

if len(parts) == 2 {
    key = parts[0]
    rawValue := parts[1] // The value is a string here
    
    // FIX: Attempt to parse the string as JSON
    var parsedValue any
    if err := json.Unmarshal([]byte(rawValue), &parsedValue); err == nil {
        // Success: The string was valid JSON
        value = parsedValue
    } else {
        // Failure: The string is a simple value (or invalid JSON)
        value = rawValue 
    }
}

ghost:
## What I did

Updated the CLI argument parsing logic for docker mcp tools call to correctly handle complex JSON values passed via the command line.
The CLI now attempts to deserialize argument values as JSON when possible, allowing objects and arrays (e.g. {}, nested maps, lists) to be passed to tools as structured data instead of raw strings.

## Related issue

Fixed #307

## What was the problem?

When calling MCP tools via the CLI, arguments provided in the form key=value were always treated as plain strings.
This caused failures for tools that expect structured JSON objects, for example:

docker mcp tools call API-patch-page \
  page_id='<UUID>' \
  properties='{}'

Instead of receiving an empty object ({}), the tool received the literal string "{}".
Downstream APIs then attempted to interpret each character of the string as a property key, resulting in validation errors such as:

Couldn't find editable properties for these identifiers: 0,1

As a result:

- Tools requiring JSON objects or arrays could not be called from the CLI
- Even valid JSON payloads failed at runtime
- Users had no way to pass structured arguments reliably

## How this change fixes it

The argument parsing logic now attempts to unmarshal each key=value argument as JSON.

Specifically:

- Each argument value is first treated as a raw string
- The CLI attempts to json.Unmarshal the value
- If unmarshalling succeeds, the parsed JSON object (map, slice, number, boolean, etc.) is used
- If it fails, the value is safely kept as a plain string
- Repeated keys continue to be aggregated into slices as before

This allows tools to receive structured data exactly as expected, without breaking existing behavior for simple string arguments.

### Before

- All CLI arguments were passed as strings
- JSON objects like {} or nested payloads were not supported
- Tools failed with validation errors when expecting structured input

### After

- JSON objects and arrays are correctly parsed and passed as structured values
- Simple string arguments continue to work unchanged
- Tools that require complex inputs can now be invoked reliably from the CLI

### How this change fixes it

The CLI argument parsing logic was updated to intelligently interpret argument values as JSON when possible.
In parseArgs, each key=value argument is now processed as follows:

- The argument is split once into key and rawValue
- The value is optimistically passed through json.Unmarshal
- If unmarshalling succeeds, the parsed value (object, array, number, boolean, etc.) is used
- If unmarshalling fails, the value is treated as a plain string
- Flag-style arguments without an explicit value continue to be supported
-  Repeated keys are aggregated into slices, preserving existing behavior

This allows users to pass structured JSON payloads (such as {}, nested objects, or arrays) directly via the CLI, while remaining fully backward-compatible with simple string arguments.

image

### DIFF
--- a/cmd/docker-mcp/tools/call.go
+++ b/cmd/docker-mcp/tools/call.go
@@ -2,6 +2,7 @@ package tools
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
@@ -70,6 +71,11 @@ func Call(ctx context.Context, version string, gatewayArgs []string, debug bool,
 	return nil
 }
 
+// toText flattens a CallToolResult into a human-readable string for CLI output.
+//
+// TextContent entries are rendered using their raw text, while any non-text
+// content types fall back to a generic fmt.Sprintf representation. Multiple
+// content blocks are joined with newlines to preserve ordering.
 func toText(response *mcp.CallToolResult) string {
 	var contents []string
 
@@ -84,6 +90,15 @@ func toText(response *mcp.CallToolResult) string {
 	return strings.Join(contents, "\n")
 }
 
+// parseArgs converts CLI arguments in the form key=value into a map suitable
+// for MCP tool invocation.
+//
+// It supports simple values as well as complex JSON payloads. When a value
+// looks like valid JSON (objects, arrays, numbers, booleans, or null), it is
+// automatically unmarshaled into its corresponding Go type. Otherwise, the
+// value is treated as a plain string.
+//
+// Repeated keys are aggregated into a slice to preserve all provided values.
 func parseArgs(args []string) map[string]any {
 	parsed := map[string]any{}
 
@@ -93,15 +108,28 @@ func parseArgs(args []string) map[string]any {
 			value any
 		)
 
+		// Split argument into key=value (only once)
 		parts := strings.SplitN(arg, "=", 2)
 		if len(parts) == 2 {
 			key = parts[0]
-			value = parts[1]
+			rawValue := parts[1]
+
+			// Attempt to parse the value as JSON.
+			// If successful, use the parsed object (map/slice/etc).
+			// Otherwise, fall back to treating it as a plain string.
+			var parsedValue any
+			if err := json.Unmarshal([]byte(rawValue), &parsedValue); err == nil {
+				value = parsedValue
+			} else {
+				value = rawValue
+			}
 		} else {
+			// Flag-style argument without an explicit value
 			key = arg
 			value = nil
 		}
 
+		// Handle repeated keys by aggregating values into a slice
 		if previous, found := parsed[key]; found {
 			switch previous := previous.(type) {
 			case []any:


cutecatfann (approved):
Good catch!

PR #371, opened 2026-01-29
fix: oauth provider registration for workingsets

bobbyhouse:
**What I did**
Implement RegisterProviderWithSnapshot for DCR provider registration. Previous implementation use the workingset server name to do a lookup in the file-based catalog.

This is unnecessary because the server snapshot data contains the provider. Additionally, this would at times cause errors when the file-based catalog was out of sync with the oci based catalog used with the profiles feature flag enabled.

**Related issue**


saucow (approved):
lgtm from perspective of similar flow as `RegisterProviderForLazySetup` but with snapshot info

PR #374, opened 2026-01-31
fix: error looking for database down migrations

bobbyhouse:
This fixes the error where the database is ahead of the migrations, but the migrations are not found. The issue occurs when the database is migrated to a new version and then you use an older version of the the application. The older version references a db with a version that is newer than the migrations it knows about.

The solution is to check the database version against the migration files available. If the database is ahead of the migration we log a warning.

Running a command such as `docker mcp profile list` would produce an error such as
```
failed to run migrations: no migration found for version 8: read down for version 8 migrations: file does not exist
```
We do not run down migrations. It appears that [golang-migrate](https://github.com/golang-migrate/migrate) will attempt to read downs if you're database version is ahead of the ups that it has available.

**What I did**

Updated migration logic to include checking to see if the database is ahead of the application. If it is, log a warning but return 0. This could also be intercepted by DD to display an informational message to the user in DD.

Log a warning
```
Warning database version 8 (/Users/bobby/.docker/mcp/mcp-toolkit.db) is newer than expected. Upgrade to the newest version to prevent issues.
```

**Related issue**

### DIFF
--- a/pkg/db/migration_test.go
+++ b/pkg/db/migration_test.go
@@ -12,7 +12,6 @@ import (
 	"github.com/golang-migrate/migrate/v4"
 	msqlite "github.com/golang-migrate/migrate/v4/database/sqlite"
 	"github.com/golang-migrate/migrate/v4/source/iofs"
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/db/migration_test.go
```diff
@@ -0,0 +1,254 @@
...
+		WithDatabaseFile(dbFile),
+		WithMigrations(createLimitedMigrationsFS(t), "."),
+	)
+	require.Error(t, err, "should error when database is ahead of migration files")
+
+	// Verify error message is descriptive
+	assert.Contains(t, err.Error(), "database version 3", "should mention current database version")
+	assert.Contains(t, err.Error(), "ahead of the current application version", "should clearly state the problem")
+	assert.Contains(t, err.Error(), dbFile, "should include database file path")
+	assert.Contains(t, err.Error(), "upgrade to the latest version", "should suggest upgrade")
```
cmrigney: **nitpick:** This could be collapsed to one check.

cmrigney (dismissed):
lgtm. Just left optional nits.

PR #389, opened 2026-02-08
Add catalog server management

slimslenderslacks:
Add server add/remove commands for catalogs to match profile syntax:
- docker mcp catalog server add <oci-ref> --server <ref1> --server <ref2>
- docker mcp catalog server remove <oci-ref> --name <name1> --name <name2>

These commands follow the same patterns as profile server commands, allowing users to manage servers in catalogs using the same familiar syntax. Servers can be added from catalog://, docker://, https://, and file:// URI schemes.

> Note that there may be some controversy here. Our initial
implementation did not include methods to update catalogs after initial creation. When a catalog is updated, it oci artifact will always have a new digest so the artifact itself is immutable if referenced by digest. But new versions of catalogs should also be easy to push (usually with new tags), and these can be used to manage profile update lifecycles.

This PR also has an update to file:// resolution

Enhance file:// JSON resolution to support MCP
This PR also has an update to file:// resolutionRegistry v0 formats:
- v0.ServerResponse (full registry response with server and _meta)
- v0.ServerJSON (server.json schema, gets wrapped in ServerResponse)
- catalog.Server (legacy Docker format, existing fallback)

Both v0 formats are converted using ConvertRegistryServerToCatalog, enabling direct use of MCP Registry standard format files with file:// URIs in both profile and catalog server commands.

### DIFF
--- a/pkg/catalog_next/tag_test.go
+++ b/pkg/catalog_next/tag_test.go
@@ -1,7 +1,6 @@
 package catalognext
 
 import (
-	"context"
 	"testing"
 
 	"github.com/stretchr/testify/assert"


pkg/workingset/workingset.go
```diff
@@ -441,12 +441,37 @@ func ResolveFile(value string) ([]Server, error) {
...
+				// Successfully parsed as v0.ServerResponse
+				catalogServer, err := ConvertRegistryServerToCatalog(&serverResp)
+				if err != nil {
+					return nil, fmt.Errorf("failed to convert v0.ServerResponse to catalog.Server: %w", err)
+				}
+				servers = []catalog.Server{catalogServer}
+			} else {
+				// Try to parse as v0.ServerJSON and wrap it
+				var serverJSON v0.ServerJSON
+				if err := json.Unmarshal(buf, &serverJSON); err == nil && serverJSON.Name != "" {
```
cmrigney: **issue:** I think this will lead to a false positive because the `catalog.Server` type also has `name` field. We'll need a more robust check here.

cmrigney (changes requested):
Generally looks good! But I think there is one issue/regression here.

PR #394, opened 2026-02-10
feat: Add id to profile create output

bobbyhouse:
**What I did**
Add output options to the profile create command so that we can return an id for the newly create profile. Supported formats are consistent with other commands (i.e. yaml and json)

**Related issue**


cmrigney (approved):
lgtm. Just left a couple chore suggestions. I could be wrong but I think we already test those cases.

PR #419, opened 2026-02-19
Fix integration tests

slimslenderslacks:
After aliases the catalog command, some integrations tests were broken

`make integration` is back working with this change.


jchangx (approved):

cmrigney (approved):
We should get these integrated into CI soon.

PR #482, opened 2026-04-15
Add npm/npx server support to MCP catalog 

cutecatfann:
**What I did**                                                                                                                                                
- Add npm/npx package support to the MCP catalog transformation pipeline, following the same pattern as the existing PyPI support (PR #392)            
- npm is the largest package type in the MCP community registry (~49% of all servers). This change adds 2,493 new importable servers to the catalog, roughly doubling the number of usable community servers.                                                                                               
- 71 npm servers are correctly skipped (no stdio transport available)                                                                                
                                                                                                                                                         
## How it works                                                                                                                                      
npm servers from the community registry are transformed into Docker catalog entries that run via `npx` inside a `node:{version}-bookworm-slim` container. The transformation:
                                                                                                                                                         
1. Resolves the Node.js version by querying the npm registry for the package's `engines.node` field (e.g., `>=18` resolves to Node 18). Defaults to Node 22 when unspecified.
2. Builds the npx command: `npx --yes [--registry URL] pkg[@version] [args...]`                                                                    
3. Mounts an npm cache volume: `docker-mcp-npm-cache-{serverName}:/root/.npm`                                                                      
4. Preserves secrets and env vars from the registry metadata                                                                                       
                                                                                                                                                         
Package priority order: OCI > PyPI > npm > Remote (unchanged, npm slots in after PyPI).                                                                
                                                                                                                                                                                                                                                                                                           
**Testing** 
Unit tests (13 transform tests + 20 parser tests, all passing)                                                                                                                 
```                                                                                                                                                                                                                                  
  === RUN   TestParseNodeVersion                                                                                                                         
      --- PASS: TestParseNodeVersion/empty_string                                                                                                      
      --- PASS: TestParseNodeVersion/greater_than_or_equal
      --- PASS: TestParseNodeVersion/greater_than_or_equal_with_patch                                                                                    
      --- PASS: TestParseNodeVersion/caret_constraint
      --- PASS: TestParseNodeVersion/tilde_constraint                                                                                                    
      --- PASS: TestParseNodeVersion/range_with_upper_bound                                                                                            
      --- PASS: TestParseNodeVersion/or_range                                                                                                            
      --- PASS: TestParseNodeVersion/garbage_input
      --- PASS: TestParseNodeVersion/just_a_number                                                                                                       
      ... (15 cases total)                                                                                                                             
  --- PASS: TestNodeVersionToImageTag (5 cases)                                                                                                          
  --- PASS: TestTransformNPM
  --- PASS: TestTransformNPMScopedPackage                                                                                                                
  --- PASS: TestTransformNPMWithCustomRegistry                                                                                                         
  --- PASS: TestTransformNPMWithPackageArgs                                                                                                              
  --- PASS: TestTransformNPMWithoutVersion                                                                                                               
  --- PASS: TestTransformNPMWithEnvVariables
  --- PASS: TestTransformNPMDisallowed                                                                                                                   
  --- PASS: TestTransformNPMAllowedByDefault                                                                                                           
  --- PASS: TestTransformNPMPackageNotFound                                                                                                              
  --- PASS: TestTransformNPMPackageNotFoundWithRemoteFallback
  --- PASS: TestTransformNPMPreferOCIOverNPM                                                                                                             
  --- PASS: TestTransformNPMPreferPyPIOverNPM                                                                                                          
  --- PASS: TestTransformNPMNilResolver                                                                                                                  
 ```    
Integration test: batch validation against live MCP community registry                                                                             
`TestIntegrationNPMTransformBatch` fetches 500 real npm-only servers from the MCP community registry and validates that `TransformToDocker` produces correct output for each one. Result: 450 passed, 0 failed, 50 skipped (remote transport took precedence), 100% success rate.                       
```                                                                                                                                                  
  === RUN   TestIntegrationNPMTransformBatch                                                                                                           
      Fetched 500 npm-only stdio servers across 18 pages of registry results
      === NPM Transform Batch Results ===                                                                                                                
      Total servers fetched: 500                                                                                                                         
      Passed:                450                                                                                                                         
      Failed:                0                                                                                                                           
      Skipped (remote/err):  50                                                                                                                          
      Success rate:          100.0% (450/450 of transformable servers)
  --- PASS: TestIntegrationNPMTransformBatch (4.51s)                                                                                                     
```                                                                                                                                             
Integration test: npm version resolver against live npm registry                                                                                   
`TestIntegrationNPMVersionResolver` validates Node.js version resolution against real npm packages:                                                    
```
  === RUN   TestIntegrationNPMVersionResolver                                                                                                            
      --- PASS: @anthropic-ai/claude-code@: engines.node resolved to ">=18" (will use image node:18-bookworm-slim)                                     
      --- PASS: @agenttrust/mcp-server@1.1.1: engines.node resolved to ">=18" (will use image node:18-bookworm-slim)                                     
      --- PASS: @contextlayer/mcp@0.0.3: engines.node resolved to "" (will use image node:22-bookworm-slim)                                              
      --- PASS: @nonexistent-scope/definitely-not-a-real-package@99.99.99: correctly returned not-found                                                  
```                                                                                                                                 
End-to-end: community registry import with `--include-npm`                                                                                         
```                                                                                                                                    
  $ docker mcp catalog create test-npm                                                                                                                   
      --from-community-registry registry.modelcontextprotocol.io                                                                                         
      --include-npm --exclude "docker/*"                                                                                                                 
                                                                                                                                                         
  Fetched 4881 servers from registry.modelcontextprotocol.io                                                                                             
    Total in registry: 6114                                                                                                                              
    Imported:          4881                                                                                                                              
      OCI (stdio):     192                                                                                                                               
      npm (stdio):     2493
      Remote:          2196                                                                                                                              
    Skipped:           1233                                                                                                                            
      pypi:            906                                                                                                                               
      no packages:     122                                                                                                                               
      mcpb:            94
      npm:             71                                                                                                                                
      nuget:           24                                                                                                                              
      oci:             16                                                                                                                                
```                                                                                                                                                     
End-to-end: single npm server profile creation and inspection
```                                                                                                                                                 
  $ docker mcp profile create --name test-npm-server                                                                                                     
      --server "https://registry.modelcontextprotocol.io/v0/servers/ai.agenttrust%2Fmcp-server/versions/1.1.1"                                           
                                                                                                                                                         
  $ docker mcp profile show test_npm_server --format yaml                                                                                                
servers:
    - type: registry
      secrets: default
      source: https://registry.modelcontextprotocol.io/v0/servers/ai.agenttrust%2Fmcp-server/versions/1.1.1
      snapshot:
        server:
            name: ai-agenttrust-mcp-server
            type: server
            image: node:18-bookworm-slim
            description: Identity, trust, and A2A orchestration for autonomous AI agents. Official A2A partner.
            title: AgentTrust — Identity & Trust for A2A Agents
            icon: https://agenttrust.ai/icon.png
            longLived: true
            remote: {}
            secrets:
                - name: ai-agenttrust-mcp-server.AGENTTRUST_API_KEY
                  env: AGENTTRUST_API_KEY
            command:
                - npx
                - --yes
                - '@agenttrust/mcp-server@1.1.1'
            volumes:
                - docker-mcp-npm-cache-ai-agenttrust-mcp-server:/root/.npm
            metadata:
                registryUrl: https://registry.modelcontextprotocol.io/v0/servers/ai.agenttrust%2Fmcp-server/versions/1.1.1
secrets:
    default:
        provider: docker-desktop-store
```                         
Note: Node 18 was correctly resolved from the package's `engines.node: ">=18"` field rather than using the default Node 22.                                                              
                                                                                                                                                         
🤖 Code and PR description generated in part with [Claude Code](https://claude.com/claude-code) with human input, creation, and review

### DIFF
--- a/cmd/docker-mcp/commands/catalog_next.go
+++ b/cmd/docker-mcp/commands/catalog_next.go
@@ -42,6 +42,7 @@ func createCatalogNextCommand() *cobra.Command {
 		Servers               []string
 		Exclude               []string
 		IncludePyPI           bool
+		IncludeNPM            bool
 	}
 
 	cmd := &cobra.Command{
@@ -92,6 +93,10 @@ When using --server without --from-profile, --from-legacy-catalog, or --from-com
 				return fmt.Errorf("--include-pypi can only be used when creating a catalog from a community registry")
 			}
 
+			if opts.IncludeNPM && opts.FromCommunityRegistry == "" {
+				return fmt.Errorf("--include-npm can only be used when creating a catalog from a community registry")
+			}
+
 			if len(opts.Exclude) > 0 && opts.FromCommunityRegistry == "" {
 				return fmt.Errorf("--exclude can only be used when creating a catalog from a community registry")
 			}
@@ -102,7 +107,16 @@ When using --server without --from-profile, --from-legacy-catalog, or --from-com
 			}
 			registryClient := registryapi.NewClient()
 			ociService := oci.NewService()
-			return catalognext.Create(cmd.Context(), dao, registryClient, ociService, args[0], opts.Servers, opts.FromWorkingSet, opts.FromLegacyCatalog, opts.FromCommunityRegistry, opts.Title, opts.IncludePyPI, opts.Exclude)
+			return catalognext.Create(cmd.Context(), dao, registryClient, ociService, args[0], catalognext.CreateOptions{
+				Servers:              opts.Servers,
+				WorkingSetID:         opts.FromWorkingSet,
+				LegacyCatalogURL:     opts.FromLegacyCatalog,
+				CommunityRegistryRef: opts.FromCommunityRegistry,
+				Title:                opts.Title,
+				IncludePyPI:          opts.IncludePyPI,
+				IncludeNPM:           opts.IncludeNPM,
+				ExcludeServers:       opts.Exclude,
+			})
 		},
 	}
 
@@ -116,6 +130,8 @@ When using --server without --from-profile, --from-legacy-catalog, or --from-com
 	flags.StringArrayVar(&opts.Exclude, "exclude", []string{}, "Server name to exclude from the catalog (can be specified multiple times, only valid with --from-community-registry)")
 	flags.BoolVar(&opts.IncludePyPI, "include-pypi", false, "Include PyPI servers when creating a catalog from a community registry")
 	cmd.Flags().MarkHidden("include-pypi") //nolint:errcheck
+	flags.BoolVar(&opts.IncludeNPM, "include-npm", false, "Include npm servers when creating a catalog from a community registry")
+	cmd.Flags().MarkHidden("include-npm") //nolint:errcheck
 
 	return cmd
 }
--- a/pkg/catalog/npm.go
+++ b/pkg/catalog/npm.go
@@ -33,19 +33,21 @@ type npmPackageInfo struct {
 	} `json:"engines"`
 }
 
-// NewNPMVersionResolver creates a resolver that queries the npm registry API.
-func NewNPMVersionResolver(httpClient *http.Client) NPMVersionResolver {
+const defaultNPMRegistryURL = "https://registry.npmjs.org"
+
+// NewNPMVersionResolver creates a resolver that queries the given npm registry URL.
+func NewNPMVersionResolver(httpClient *http.Client, npmRegistryURL string) NPMVersionResolver {
 	return func(ctx context.Context, identifier, version, registryBaseURL string) (string, bool) {
 		// Only query npm for standard npm registry
-		if registryBaseURL != "" && registryBaseURL != "https://registry.npmjs.org" {
+		if registryBaseURL != "" && registryBaseURL != npmRegistryURL {
 			return "", true // assume found for non-standard registries
 		}
 
 		var url string
 		if version != "" {
-			url = fmt.Sprintf("https://registry.npmjs.org/%s/%s", identifier, version)
+			url = fmt.Sprintf("%s/%s/%s", npmRegistryURL, identifier, version)
 		} else {
-			url = fmt.Sprintf("https://registry.npmjs.org/%s/latest", identifier)
+			url = fmt.Sprintf("%s/%s/latest", npmRegistryURL, identifier)
 		}
 
 		req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
@@ -78,7 +80,7 @@ func DefaultNPMVersionResolver() NPMVersionResolver {
 		Transport: desktop.ProxyTransport(),
 		Timeout:   10 * time.Second,
 	}
-	return NewNPMVersionResolver(client)
+	return NewNPMVersionResolver(client, defaultNPMRegistryURL)
 }
 
 // parseNodeVersion extracts the Node.js major version from a semver constraint.
--- a/pkg/catalog/npm.go
+++ b/pkg/catalog/npm.go
@@ -12,13 +12,14 @@ import (
 )
 
 const (
-	defaultNodeVersion = "22"
+	defaultNodeVersion = "24"
 	nodeBaseImage      = "node"
 )
 
 var (
-	nodeVersionRe     = regexp.MustCompile(`(?:>=|>|\^|~)\s*(\d+)`)
-	bareNodeVersionRe = regexp.MustCompile(`^\s*(\d+)\s*$`)
+	nodePinVersionRe    = regexp.MustCompile(`(?:\^|~)\s*(\d+)`)
+	nodeLowerBoundRe    = regexp.MustCompile(`(?:>=|>)\s*\d+`)
+	bareNodeVersionRe   = regexp.MustCompile(`^\s*(\d+)\s*$`)
 )
 
 // NPMVersionResolver resolves the minimum Node.js version for an npm package.
@@ -80,18 +81,25 @@ func DefaultNPMVersionResolver() NPMVersionResolver {
 	return NewNPMVersionResolver(client)
 }
 
-// parseNodeVersion extracts the minimum major Node.js version from a semver constraint.
-// It looks for the first >=, >, ^, or ~ operator and returns the major version number.
+// parseNodeVersion extracts the Node.js major version from a semver constraint.
+// Pinning operators (^ or ~) resolve to a specific major version.
+// >= and > mean "this or newer", so we use the latest (default).
 // A bare major version (e.g., "18") is also accepted.
-// Examples: ">=18" -> "18", "^20.0.0" -> "20", ">=16.17.0" -> "16", "18" -> "18", "" -> "" (use default)
+// Examples: "^20.0.0" -> "20", "~18" -> "18", ">=18" -> "" (use latest), ">16" -> "" (use latest),
+// "18" -> "18", "" -> "" (use default)
 func parseNodeVersion(enginesNode string) string {
 	if enginesNode == "" {
 		return ""
 	}
 
-	match := nodeVersionRe.FindStringSubmatch(enginesNode)
-	if match != nil {
+	// Check for pinning constraints (^ or ~) first - these specify a major version
+	if match := nodePinVersionRe.FindStringSubmatch(enginesNode); match != nil {
 		return match[1]
+	}
+
+	// >= and > mean "this or newer", so use the latest (default)
+	if nodeLowerBoundRe.MatchString(enginesNode) {
+		return ""
 	}
 
 	// Fall back to bare major version (e.g., "18")
--- a/pkg/catalog/npm.go
+++ b/pkg/catalog/npm.go
@@ -1,0 +1,112 @@
+package catalog
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"regexp"
+	"time"
+
+	"github.com/docker/mcp-gateway/pkg/desktop"
+)
+
+const (
+	defaultNodeVersion = "22"
+	nodeBaseImage      = "node"
+)
+
+var (
+	nodeVersionRe     = regexp.MustCompile(`(?:>=|>|\^|~)\s*(\d+)`)
+	bareNodeVersionRe = regexp.MustCompile(`^\s*(\d+)\s*$`)
+)
+
+// NPMVersionResolver resolves the minimum Node.js version for an npm package.
+// It returns the minimum Node.js major version string (e.g., "18") or empty string if unknown,
+// and a boolean indicating whether the package was found.
+type NPMVersionResolver func(ctx context.Context, identifier, version, registryBaseURL string) (string, bool)
+
+type npmPackageInfo struct {
+	Engines struct {
+		Node string `json:"node"`
+	} `json:"engines"`
+}
+
+// NewNPMVersionResolver creates a resolver that queries the npm registry API.
+func NewNPMVersionResolver(httpClient *http.Client) NPMVersionResolver {
+	return func(ctx context.Context, identifier, version, registryBaseURL string) (string, bool) {
+		// Only query npm for standard npm registry
+		if registryBaseURL != "" && registryBaseURL != "https://registry.npmjs.org" {
+			return "", true // assume found for non-standard registries
+		}
+
+		var url string
+		if version != "" {
+			url = fmt.Sprintf("https://registry.npmjs.org/%s/%s", identifier, version)
+		} else {
+			url = fmt.Sprintf("https://registry.npmjs.org/%s/latest", identifier)
+		}
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+		if err != nil {
+			return "", false
+		}
+
+		resp, err := httpClient.Do(req)
+		if err != nil {
+			return "", false
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode != http.StatusOK {
+			return "", false
+		}
+
+		var info npmPackageInfo
+		if err := json.NewDecoder(resp.Body).Decode(&info); err != nil {
+			return "", false
+		}
+
+		return parseNodeVersion(info.Engines.Node), true
+	}
+}
+
+// DefaultNPMVersionResolver creates a resolver using the default HTTP client with proxy transport.
+func DefaultNPMVersionResolver() NPMVersionResolver {
+	client := &http.Client{
+		Transport: desktop.ProxyTransport(),
+		Timeout:   10 * time.Second,
+	}
+	return NewNPMVersionResolver(client)
+}
+
+// parseNodeVersion extracts the minimum major Node.js version from a semver constraint.
+// It looks for the first >=, >, ^, or ~ operator and returns the major version number.
+// A bare major version (e.g., "18") is also accepted.
+// Examples: ">=18" -> "18", "^20.0.0" -> "20", ">=16.17.0" -> "16", "18" -> "18", "" -> "" (use default)
+func parseNodeVersion(enginesNode string) string {
+	if enginesNode == "" {
+		return ""
+	}
+
+	match := nodeVersionRe.FindStringSubmatch(enginesNode)
+	if match != nil {
+		return match[1]
+	}
+
+	// Fall back to bare major version (e.g., "18")
+	if bare := bareNodeVersionRe.FindStringSubmatch(enginesNode); bare != nil {
+		return bare[1]
+	}
+
+	return ""
+}
+
+// nodeVersionToImageTag maps a Node.js major version to the appropriate Docker image tag.
+func nodeVersionToImageTag(nodeVersion string) string {
+	if nodeVersion == "" {
+		nodeVersion = defaultNodeVersion
+	}
+
+	return fmt.Sprintf("%s:%s-bookworm-slim", nodeBaseImage, nodeVersion)
+}
--- a/pkg/catalog/npm.go
+++ b/pkg/catalog/npm.go
@@ -17,9 +17,9 @@ const (
 )
 
 var (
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

**Comment by cutecatfann** [comment]
Ran the npm transform pipeline against the entire MCP community registry to validate the implementation.                                               
```                                                                                                                                           
  Catalog Check Results (transform validation)                                                                                                           
                                                                                                                                                         
  Pages read:            62                                                                                                                              
  Total npm servers:     2691                                                                                                                          
  Passed:                2486
  Failed:                0                                                                                                                               
  Skipped (remote/err):  205
  Success rate:          100.0% (2486/2486 transformable)                                                                                                
```                                                                                                                                             

All 2,486 transformable npm servers produce correct Docker catalog output (image, command, volumes, metadata). The 205 skipped servers are npm packages that also expose a remote transport (SSE/streamable-http), so `TransformToDocker` correctly resolves them as `remote` instead of `npm` per the priority order (OCI > PyPI > npm > Remote).                                                                                                                     
```                                                                                                                                             
  Registry Breakdown (separate count script)                                                                                                             
   
  Total servers:                       6180                                                                                                              
  Has npm (any transport):             2808  (45.4%)                                                                                                   
  Has npm+stdio:                       2763  (44.7%)
  Has OCI:                              235  (3.8%)                                                                                                      
  Has PyPI:                            1030  (16.7%)
  Has remote/SSE transport:            2562  (41.5%)                                                                                                     
  npm+stdio, no OCI, no PyPI:          2694  (our filter target)                                                                                         
```
The registry now has 6,180 total servers and 2,808 npm servers. The registry has been cleaned up since the context doc was written, but the npm proportion (45.4%) is consistent with the original claim (49.5%). npm remains the largest package type.                                                                                                                                          
                                                                                                                                                       
The check script found 2,691 npm-only servers; the count script found 2,694. The difference of 3 servers is a timing artifact, the two scripts ran several hours apart against a live registry that can change between requests.

File: cmd/docker-mcp/commands/catalog_next.go [hunk 1]
```diff
@@ -102,7 +107,16 @@ When using --server without --from-profile, --from-legacy-catalog, or --from-com
 			}
 			registryClient := registryapi.NewClient()
 			ociService := oci.NewService()
-			return catalognext.Create(cmd.Context(), dao, registryClient, ociService, args[0], opts.Servers, opts.FromWorkingSet, opts.FromLegacyCatalog, opts.FromCommunityRegistry, opts.Title, opts.IncludePyPI, opts.Exclude)
+			return catalognext.Create(cmd.Context(), dao, registryClient, ociService, args[0], catalognext.CreateOptions{
```
cmrigney, line 110: **praise:** Good refactor on CreateOptions!

pkg/catalog/npm.go
```diff
@@ -0,0 +1,112 @@
...
+	"fmt"
+	"net/http"
+	"regexp"
+	"time"
+
+	"github.com/docker/mcp-gateway/pkg/desktop"
+)
+
+const (
+	defaultNodeVersion = "22"
```
cmrigney: **chore:** The current LTS is v24, we will probably want to use that instead.

pkg/catalog/npm.go
```diff
@@ -0,0 +1,112 @@
...
+		Transport: desktop.ProxyTransport(),
+		Timeout:   10 * time.Second,
+	}
+	return NewNPMVersionResolver(client)
+}
+
+// parseNodeVersion extracts the minimum major Node.js version from a semver constraint.
+// It looks for the first >=, >, ^, or ~ operator and returns the major version number.
+// A bare major version (e.g., "18") is also accepted.
+// Examples: ">=18" -> "18", "^20.0.0" -> "20", ">=16.17.0" -> "16", "18" -> "18", "" -> "" (use default)
```
cmrigney: **chore:** I think we should resolve to the latest version as possible. For example, resolving `>=18` would resolve to the latest LTS (v24).

pkg/catalog/npm.go
```diff
@@ -0,0 +1,112 @@
...
+func parseNodeVersion(enginesNode string) string {
```
cmrigney, line 92: _From Claude:_

**nitpick (non-blocking)**: `>N` and `>=N` produce the same result

The regex captures the numeric major that follows a `>`, `>=`, `^`, or `~` operator. Semantically `>18` means "strictly greater than any 18.x.x" so the minimum Node is 19, but `parseNodeVersion(">18")` returns `"18"`. In practice almost no package author writes `>N` without a patch, so this is very unlikely to matter — but worth a line of handling if you want strict correctness.

pkg/catalog/npm_integration_test.go
```diff
@@ -0,0 +1,311 @@
...
+// registry and validates that TransformToDocker produces correct output for each one.
+// This test exercises the full npm transformation pipeline against live registry data.
+//
+// Run with: go test -count=1 ./pkg/catalog/... -run TestIntegrationNPMTransformBatch -v -timeout 5m
+func TestIntegrationNPMTransformBatch(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping integration test in short mode")
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
cmrigney: **nitpick (non-blocking):**


```suggestion
	ctx, cancel := context.WithTimeout(t.Context(), 5*time.Minute)
```

pkg/catalog/npm_integration_test.go
```diff
@@ -0,0 +1,311 @@
...
+	if failed > 0 {
+		t.Errorf("%d servers failed transformation validation", failed)
+	}
+
+	if passed == 0 {
+		t.Error("No servers passed npm transformation - something is wrong")
+	}
+
+	successRate := float64(passed) / float64(passed+failed) * 100
+	t.Logf("Success rate:          %.1f%% (%d/%d of transformable servers)", successRate, passed, passed+failed)
```
cmrigney: **question:** Is this redundant since success rate will always be 100% by this point?

pkg/catalog/npm_integration_test.go
```diff
@@ -0,0 +1,311 @@
...
+
+type serverResponseJSON struct {
+	Server ServerDetail `json:"server"`
+}
+
+// fetchNPMOnlyServers fetches npm-only servers from the MCP community registry.
+// It paginates through the registry collecting servers that have only npm packages
+// with stdio transport. Stops after maxServers are collected or maxPages are read.
+// Returns partial results with a non-nil error if a page fetch fails.
+func fetchNPMOnlyServers(ctx context.Context, client *http.Client, maxServers int, maxPages int) ([]serverResponseJSON, int, error) {
```
cmrigney: **thought:** This makes our integration tests dependent on an external service so we have no way to fix them if the service goes down or gets changed (and I'm not sure if integration tests block any workflows). What I might propose instead is downloading the data you need into a `testdata` folder and reading that, essentially simulating this fetch. wdyt?

pkg/catalog/registry_to_catalog.go
```diff
@@ -50,6 +53,22 @@ func WithPyPIResolver(resolver PyPIVersionResolver) TransformOption {
 	}
 }
 
+// WithAllowNPM controls whether npm packages are considered during transformation.
+// By default, npm packages are allowed.
```
cmrigney: Maybe typo here and on pypi?

```suggestion
// By default, npm packages are not allowed.
```

cmrigney (changes requested):
Tested and works for me! The only blockers are the chores. I think those are important for security reasons. Everything else is just a suggestion.

pkg/catalog/registry_to_catalog.go
cutecatfann: Hahaha! Whoops haha

pkg/catalog/npm.go
cutecatfann: Good call, updated defaultNodeVersion to "24"

pkg/catalog/npm.go
cutecatfann: Agreed. Updated `parseNodeVersion` to match the PyPI pattern: `>=` and `>` now return `""` (use latest default), while `^` and `~` still pin to the specified major version. So >=18 will now resolve to `node:24-bookworm-slim` instead of `node:18-bookworm-slim`

pkg/catalog/npm.go
cutecatfann: Fixed!

pkg/catalog/npm_integration_test.go
cutecatfann: Fixed!

pkg/catalog/npm_integration_test.go
cutecatfann: You're right, removed it. The passed/failed counts above already convey the same information

pkg/catalog/npm_integration_test.go
cutecatfann: Good idea. Replaced both integration tests with `testdata-backed` versions. Created `pkg/catalog/testdata/` with representative npm registry responses, and the tests now use `httptest.NewServer` to serve them locally. No more external service dependency.

pkg/catalog/registry_to_catalog.go
cutecatfann: Fixed!

pkg/catalog_next/create.go
```diff
@@ -218,15 +231,20 @@ func createCatalogFromCommunityRegistry(ctx context.Context, registryClient regi
...
 			skippedByType["excluded"]++
 			continue
 		}
 
-		catalogServer, transformSource, err := legacycatalog.TransformToDocker(ctx, serverResp.Server, legacycatalog.WithAllowPyPI(includePyPI), legacycatalog.WithPyPIResolver(legacycatalog.DefaultPyPIVersionResolver()))
+		catalogServer, transformSource, err := legacycatalog.TransformToDocker(ctx, serverResp.Server,
+			legacycatalog.WithAllowPyPI(includePyPI),
+			legacycatalog.WithPyPIResolver(legacycatalog.DefaultPyPIVersionResolver()),
+			legacycatalog.WithAllowNPM(includeNPM),
+			legacycatalog.WithNPMResolver(legacycatalog.DefaultNPMVersionResolver()),
```
jchangx: Each npm lookup can block up to 10s, and this loop is sequential — one stuck request stalls the rest of a ~2,500-server import. Wrapping the outer loop in an `errgroup` with a bounded worker pool (8–16 feels about right) would make bulk imports much more predictable. PyPI would benefit from the same treatment, so it might be worth extracting a shared helper.

pkg/catalog/npm_integration_test.go
```diff
@@ -0,0 +1,258 @@
...
+			return
+		}
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write(data)
+	}))
+	defer srv.Close()
+
+	// Create a resolver that points to our test server instead of the real npm registry.
+	// We construct the resolver manually rather than using NewNPMVersionResolver because
+	// the standard resolver hardcodes the npm registry URL.
```
jchangx: The test reimplements ~30 lines of `NewNPMVersionResolver` because the real one hardcodes the registry URL. If you add a `baseURL string` parameter to `NewNPMVersionResolver` and have `DefaultNPMVersionResolver` pass `https://registry.npmjs.org`, the test can call the production function directly — no duplication, and no risk of the two copies drifting.

pkg/catalog/registry_to_catalog.go
```diff
@@ -230,6 +249,36 @@ func extractPyPIInfo(pkg model.Package, pythonVersion string, serverName string)
...
+	// Add package identifier with optional version
+	if pkg.Version != "" {
+		command = append(command, fmt.Sprintf("%s@%s", pkg.Identifier, pkg.Version))
+	} else {
+		command = append(command, pkg.Identifier)
+	}
+
+	// Add npm cache volume (keyed per server to avoid cross-contamination)
+	volumeName := fmt.Sprintf("docker-mcp-npm-cache-%s", serverName)
+	volumes = []string{volumeName + ":/root/.npm"}
```
jchangx, line 294: Cache is mounted at `/root/.npm`, but `extractUserFromRuntimeArgs` can override the user. A non-root user can't write there, so the cache silently does nothing and every run re-downloads. Simplest fix: when a user override is present, either point the volume at that user's home (`/home/<user>/.npm`) or skip the cache volume entirely. PyPI has the same shape with `/root/.cache/uv` — worth fixing together.

pkg/catalog/testdata/npm_registry_servers.json
```diff
@@ -0,0 +1,254 @@
+{
```
jchangx, line 1: 10 servers vs the old 500 is a big coverage drop. Moving to testdata was the right call — suggest re-capturing a one-time snapshot of ~100 that deliberately covers the interesting shapes: scoped packages, custom `registryBaseUrl`, packages with env vars and package args, remote fallbacks, and a spread of `engines.node` (empty, pin, lower bound, bare number). That keeps the breadth without the flakiness.

pkg/catalog/registry_to_catalog.go
```diff
@@ -230,6 +249,36 @@ func extractPyPIInfo(pkg model.Package, pythonVersion string, serverName string)
...
+
+	// Add package identifier with optional version
+	if pkg.Version != "" {
+		command = append(command, fmt.Sprintf("%s@%s", pkg.Identifier, pkg.Version))
+	} else {
+		command = append(command, pkg.Identifier)
+	}
+
+	// Add npm cache volume (keyed per server to avoid cross-contamination)
+	volumeName := fmt.Sprintf("docker-mcp-npm-cache-%s", serverName)
```
jchangx, line 293: `extractServerName` strips `/` and `.`, but Docker volume names must match `[a-zA-Z0-9][a-zA-Z0-9_.-]*`. A server name with `@`, a space, or a leading hyphen will fail at container start. Safest fix is a slug pass in `extractServerName` — e.g., replace any non-matching character with `-` and trim leading non-alphanumerics — so the fix covers volume names, secret names, and the config schema name in one place.

jchangx (commented):
A few additional suggestions on top of the existing thread — all constructive, none are blockers.

pkg/catalog_next/create.go
cutecatfann: Agreed! Wrapped the transform loop in an `errgroup` with `SetLimit(12)`. Resolvers are now created once and shared. Results use a position-indexed slice to preserve ordering without a post-sort. Also applies to PyPI since they share the same loop

pkg/catalog/npm_integration_test.go
cutecatfann: Good catch. Added a `npmRegistryURL` parameter to `NewNPMVersionResolver`, and `DefaultNPMVersionResolver` passes the default. The test now calls `NewNPMVersionResolver(srv.Client(), srv.URL)` directly

pkg/catalog/registry_to_catalog.go
cutecatfann: Fixed. After a user override is applied, we now filter out any volumes targeting `/root/`. This covers both npm (`/root/.npm`) and PyPI (`/root/.cache/uv`). Went with the "skip the cache volume" approach since guessing non-root home directories is fragile

pkg/catalog/testdata/npm_registry_servers.json
cutecatfann: Extended to 100!

pkg/catalog/registry_to_catalog.go
cutecatfann: Added a slug pass in `extractServerName`. It now replaces any character outside `[a-zA-Z0-9_-]` with `- `and trims leading non-alphanumeric characters. So `@scope/pkg` becomes `scope-pkg`. Backward compatible with existing names since `/` and `.` were already replaced with `-`

pkg/catalog/npm_test.go
```diff
@@ -11,20 +11,20 @@ func TestParseNodeVersion(t *testing.T) {
...
+		{"greater than or equal with patch", ">=16.17.0", ""},
+		{"greater than or equal with minor", ">=20.0", ""},
 		{"caret constraint", "^18", "18"},
 		{"caret with patch", "^20.0.0", "20"},
 		{"tilde constraint", "~18", "18"},
 		{"tilde with minor", "~20.11", "20"},
-		{"greater than", ">16", "16"},
-		{"range with upper bound", ">=18 <22", "18"},
+		{"greater than", ">16", ""},
+		{"range with upper bound", ">=18 <22", ""},
```
cmrigney, line 22: **nitpick:** This should technically be 22.

pkg/catalog/npm_test.go
```diff
@@ -11,20 +11,20 @@ func TestParseNodeVersion(t *testing.T) {
...
 		{"or range", "18 || 20 || 22", ""},
 		{"garbage input", "foobar", ""},
 		{"just a number", "18", "18"},
-		{"with spaces", ">= 18", "18"},
-		{"complex range", ">=18.0.0 <25.0.0", "18"},
+		{"with spaces", ">= 18", ""},
+		{"complex range", ">=18.0.0 <25.0.0", ""},
```
cmrigney, line 27: **nitpick:** This should technically be 25.

pkg/catalog/registry_to_catalog.go
```diff
@@ -32,10 +33,12 @@ type TransformOption func(*transformOptions)
 type transformOptions struct {
 	allowPyPI    bool
 	pypiResolver PyPIVersionResolver
+	allowNPM     bool
+	npmResolver  NPMVersionResolver
 }
 
 // WithAllowPyPI controls whether PyPI packages are considered during transformation.
-// By default, PyPI packages are allowed.
+// By default, PyPI packages are not allowed.
```
cmrigney, line 41: My AI model is telling me I was wrong about this. The default _is_ allowed after all lol. Sorry!

cmrigney (approved):
Looks good, just left a couple optional comments.

PR #487, opened 2026-05-04
Removing McpGatewayOAuth feature flag

cutecatfann:
**What I did**
Removing fully released McpGatewayOAuth feature flag and depreciating old code paths


cmd/docker-mcp/oauth/auth.go
```diff
@@ -338,9 +338,6 @@ func authorizeCommunityMode(ctx context.Context, serverName string, scopes strin
 	// token is safely stored in docker pass. We defer this until after storage
 	// succeeds so that if the authorize flow fails at any earlier step, the
 	// user retains their existing Desktop authorization as a fallback.
-	// The docker-desktop-mcp-oauth plugin (pattern docker/mcp/oauth/**) has
-	// higher priority than docker-pass (**), so stale Desktop entries would
-	// shadow the fresh token if not removed.
```
austin5456, line 343: is this no longer the case?

austin5456 (approved):

cmd/docker-mcp/oauth/auth.go
cutecatfann: Still the case!

PR #489, opened 2026-05-16
fix: reuse containers per session

### LINKED ISSUE #483: Gateway leaks containers and zombie processes under concurrent tool calls
## Description

The `docker mcp gateway` (v0.34.0) leaks containers and accumulates zombie processes when handling concurrent MCP tool calls from multiple background agents. A single Claude Code session running 10 parallel agents produced **180+ orphaned containers** and **526 zombie processes**, bringing the host machine to 100% CPU.

## Environment

- **docker-mcp**: v0.34.0
- **Docker Desktop**: macOS (Darwin 25.3.0)
- **Client**: Claude Code CLI v2.1.110
- **MCP servers registered**: 9 (github-official, brave, fetch, playwright, context7, etc.)

## Steps to reproduce

1. Connect `docker mcp` to Claude Code (`docker mcp client connect claude-code`)
2. Register multiple MCP servers including `github-official`, `brave`, `fetch`
3. Start a Claude Code session
4. Launch **10+ background agents** that make concurrent MCP tool calls (`brave_web_search`, `fetch`) through the gateway

In my case this was triggered by a deep-research workflow dispatching 10 parallel agents, each making web searches and URL fetches simultaneously through the single gateway process.

### Container leak
The gateway creates a new Docker container per tool call but **does not reuse or clean up** containers under concurrent load. Over ~2 hours:
- **180+ running containers** of `ghcr.io/github/github-mcp-server` accumulated
- Containers stay in "Up" state indefinitely
- New containers keep being created even after the calling agents have finished

### Zombie processes
The gateway process does not reap child processes:
```
Gateway PID 82681: 1475 open FDs, 726 children, 526 zombies
```

For comparison, 8 other gateway instances on the same machine (handling normal single-agent load) had **0 children and 0 zombies** each:
```
Gateway 82681 (10 concurrent agents) : 1475 FDs, 726 children, 526 zombies
Gateway 2922  (normal usage)         :   22 FDs,   0 children,   0 zombies
```

### Resource impact
- Docker VM: **332% CPU** (saturating 3+ cores)
- `kernel_task`: 62% CPU (thermal throttling)
- Host machine essentially unusable (idle < 3%)

### Self-reinforcing loop
After killing all containers, the gateway **immediately respawns** new ones (~20 containers in 10 seconds). The only fix was to `kill` the gateway process itself.

## Expected behavior

1. The gateway should **reuse containers** across tool calls to the same MCP server, or implement a container pool
2. Child processes should be properly reaped (no zombies)
3. There should be a **max concurrent containers** limit
4. Containers should be cleaned up when the calling session/agent finishes
5. When the parent Claude Code session is idle (no pending tool calls), the gateway should not spawn new containers

## Diagnostic data

**Container labels** (from `docker run` command observed via `ps`):
```
docker run --rm -i --init --security-opt no-new-privileges \
  --cpus 1 --memory 2Gb --pull never \
  -l docker-mcp=true \
  -l docker-mcp-tool-type=mcp \
  -l docker-mcp-name=github-official \
  -l docker-mcp-transport=stdio \
  -e GITHUB_PERSONAL_ACCESS_TOKEN \
  ghcr.io/github/github-mcp-server@sha256:5049daf...
```

**Container creation rate** (from `docker ps` timestamps):
```
15:11  — 39 containers created in 1 minute
15:12  — 68 containers created in 1 minute
15:13  — 11 containers
16:00  — 9 containers
16:01  — 22 containers
... continuous creation for 2+ hours
```

**Process tree**:
```
cmux → login → zsh → claude (PID 82364) → docker mcp gateway run (PID 82681)
                                            ├── 726 child docker-run processes
                                            └── 526 zombie processes
```

## Related

- #315 — similar container cleanup issue with Playwright MCP server (closed)

## Workaround

Kill the affected gateway process (`kill <gateway-pid>`). The Claude Code session survives and will spawn a fresh gateway on next MCP tool call. Then clean up orphaned containers:
```bash
docker rm -f $(docker ps -aq --filter "label=docker-mcp-name=github-official")
```

areebahmeddd:
**What I did**

* Reuse a single Docker MCP container per client session instead of creating a new container for every tool call, preventing container leaks and zombie processes under concurrent load
* Added safer concurrent client initialization and automatic cleanup of cached containers when client sessions disconnect
* Used VS Code Copilot (Claude Sonnet 4.6) to understand the issue context and help generate a fix. Changes were self-reviewed

**Related issue**

Fixes #483

**A picture of a cute animal**

image

### DIFF
--- a/pkg/gateway/clientpool.go
+++ b/pkg/gateway/clientpool.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"strings"
 	"sync"
+	"sync/atomic"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 
@@ -73,17 +74,11 @@ func (cp *clientPool) longLived(serverConfig *catalog.ServerConfig, config *clie
 		return false
 	}
 
-	// Remote servers are always treated as long-lived since no container is involved
+	// Remote servers are always long-lived
 	if serverConfig.IsRemote() {
 		return true
 	}
 
-	// One container per (server, session), not per call
-	if serverConfig.Spec.Image != "" {
-		return true
-	}
-
-	// For Docker-based servers, respect the LongLived flag
 	return serverConfig.Spec.LongLived || cp.LongLived
 }
 
@@ -194,7 +189,10 @@ func (cp *clientPool) ReleaseClientsForSession(session *mcp.ServerSession) {
 	cp.clientLock.Unlock()
 
 	for _, kc := range toClose {
-		client, err := kc.Getter.GetClient(context.TODO()) // should be cached
+		if !kc.Getter.started.Load() {
+			continue // GetClient never called; no container to stop
+		}
+		client, err := kc.Getter.GetClient(context.Background()) // should be cached
 		if err == nil {
 			client.Session().Close()
 		}
@@ -322,7 +320,8 @@ func (cp *clientPool) baseArgs(name string) []string {
 	}
 
 	// Add a few labels to the container for identification
-	args = append(args,
+	args = append(
+		args,
 		"-l", "docker-mcp=true",
 		"-l", "docker-mcp-tool-type=mcp",
 		"-l", "docker-mcp-name="+name,
@@ -451,9 +450,10 @@ func maskSecret(value string) string {
 }
 
 type clientGetter struct {
-	once   sync.Once
-	client mcpclient.Client
-	err    error
+	once    sync.Once
+	started atomic.Bool
+	client  mcpclient.Client
+	err     error
 
 	serverConfig *catalog.ServerConfig
 	cp           *clientPool
@@ -474,6 +474,7 @@ func (cg *clientGetter) IsClient(client mcpclient.Client) bool {
 }
 
 func (cg *clientGetter) GetClient(ctx context.Context) (mcpclient.Client, error) {
+	cg.started.Store(true)
 	cg.once.Do(func() {
 		createClient := func() (mcpclient.Client, error) {
 			cleanup := func(context.Context) error { return nil }
--- a/pkg/gateway/clientpool.go
+++ b/pkg/gateway/clientpool.go
@@ -78,6 +78,11 @@ func (cp *clientPool) longLived(serverConfig *catalog.ServerConfig, config *clie
 		return true
 	}
 
+	// One container per (server, session), not per call
+	if serverConfig.Spec.Image != "" {
+		return true
+	}
+
 	// For Docker-based servers, respect the LongLived flag
 	return serverConfig.Spec.LongLived || cp.LongLived
 }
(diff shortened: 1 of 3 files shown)


areebahmeddd:
gentle ping! @cutecatfann 🙂

cutecatfann (changes requested):
Thanks for the PR and for digging into this issue. Looks like the right start :)                                                               
                                                                                                                             
A few things I'd like addressed before merging:                                                                            
                                                                                                                             
1. The `Spec.Image != ""` override in longLived() is too broad                                                               
```                                                                                                                       
  if serverConfig.Spec.Image != "" {                                                                                         
      return true                                                                                                          
  }
```
This forces every container-based server to be long-lived regardless of its LongLived flag. Some servers are intentionally stateless and expect a fresh container per call. The double-checked locking alone already fixes the duplication race (multiple goroutines creating containers for the same key concurrently). Please remove this addition, the existing `serverConfig.Spec.LongLived || cp.LongLived` logic should continue to control caching behavior, and the race fix ensures that when caching IS enabled, only one container is created.

2. The concurrency test doesn't test the actual code path                                                                  
`TestAcquireClientNoDuplicatesUnderConcurrency` manually reimplements the locking logic inside each goroutine instead of calling `cp.AcquireClient()`. This means the test would pass even if AcquireClient itself still had the race. Please rewrite it to call the real method (you may need to mock the Docker client / container creation to avoid starting real containers).
                                                                                                                           
3. ReleaseClientsForSession can trigger container creation                                                                 
If a clientGetter was stored in the map but GetClient was never called (container never started), calling `kc.Getter.GetClient(context.TODO())` inside ReleaseClientsForSession will actually trigger the sync.Once and start a container just to close it                                                          

4. Minor: context.TODO()
`Both ReleaseClientsForSession` and the existing Close() use context.TODO(). For the new code, consider accepting or deriving a proper context so the cleanup calls can be cancelled during gateway shutdown.

areebahmeddd:
> The Spec.Image != "" override in longLived() is too broad

cool, i've removed the image check entirely

> The concurrency test doesn't test the actual code path

re wrote the test to use cp.AcquireClient() directly now

> ReleaseClientsForSession can trigger container creation

let me add a check so we don’t start a container during cleanup

> Minor: context.TODO()

sure 👍🏻

cutecatfann (dismissed):

cutecatfann:
Just FYI looks like you will have to rebase off of https://github.com/docker/mcp-gateway/pull/494 to be permitted to merge. Just @ me when the commit is up

areebahmeddd:
rebased! cc @cutecatfann

cutecatfann:
Thanks for your patience with the CI! It was being funny for external contributors. Looks like your most recent commit was unverified FYI

areebahmeddd:
Hmm, weird. i do use gpg signing. let me try something

austin5456 (approved):

PR #493, opened 2026-05-21
Reduce insecurity of secret setting command in documentation

tjnicholas:
The existing example encourages an unsafe pattern where the a secret value is written to disk in clear text - for no reason at all. Given the purpose of the feature is secure storage of secrets, this seems worth avoiding.

### DIFF
--- a/docs/generator/reference/docker_mcp_secret_set.yaml
+++ b/docs/generator/reference/docker_mcp_secret_set.yaml
@@ -18,8 +18,7 @@ examples: |-
     ### Pass the secret via STDIN
 
     ```console
-    echo my-secret-password > pwd.txt
-    cat pwd.txt | docker mcp secret set postgres_password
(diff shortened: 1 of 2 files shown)


cutecatfann:
Thank you for catching!

cutecatfann (dismissed):

tjnicholas:
Updated in a second location. Same change as in the yaml file. 

I'm not 100% sure this is why the validate check had failed, but it was incorrect anyway. 

Thanks for the approval @cutecatfann, sorry I had to invalidate it with a fresh commit!

cutecatfann (approved):

cutecatfann:
Looks like you are missing a verified signature on your PR :)

tjnicholas:
Well, that was more frustrating to fix than I'd expected. 

But commits are now verified!

PR #497, opened 2026-05-27
test(secretsscan): assemble token-shaped inputs at runtime

tuna-docker:
## Summary

The two literals in `pkg/secretsscan/secrets_test.go` `TestContainsSecrets` matched the GitHub-PAT and Docker-PAT regex shapes closely enough that GitHub's secret-scanning service flags the file as containing leaked credentials. Push protection also blocks any future push containing a similarly-shaped literal.

This change builds the test inputs at runtime from a literal prefix (`ghp_`, `dckr_pat_`) plus repeated padding via `strings.Repeat`, so the source code no longer contains a complete-shape token. `ContainsSecrets` sees the assembled string and continues to return `true` for both patterns.

## Test

```
$ go test ./pkg/secretsscan/ -count=1 -v
=== RUN   TestDoesntContainsSecrets
--- PASS: TestDoesntContainsSecrets (0.00s)
=== RUN   TestContainsSecrets
--- PASS: TestContainsSecrets (0.00s)
PASS
```

### DIFF
--- a/pkg/secretsscan/secrets_test.go
+++ b/pkg/secretsscan/secrets_test.go
@@ -1,6 +1,7 @@
 package secretsscan
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


cutecatfann (approved):
Thank you! I didn't have these flagging before, so interesting they started flagging now

PR #499, opened 2026-05-30
Refuse long-lived server startup with missing secrets

jstar0:
## Summary

Fixes #433.

Long-lived local MCP server containers can keep running with placeholder secret values if they are started before required secrets are available. Once that container is cached, later calls can continue using the stale environment even after the user stores the real secret.

This change makes local long-lived Docker servers require verified required secrets before a client/container is created. If the secrets engine cannot be queried, fallback `se://` URIs are not synthesized for those long-lived local servers, so missing secrets remain visible and the gateway returns a clear error instead of caching a container with a placeholder value.

## Changes

- Skip unverified fallback secret URIs for local long-lived Docker servers.
- Avoid backfilling those fallback URIs through another server that declares the same secret name.
- Return an error before creating or caching a long-lived client when required secrets are missing.
- Treat empty required secret values as missing for local long-lived Docker servers.
- Keep failed `mcp-add` validation attempts from enabling the server in memory.
- Preserve existing fallback/placeholder behavior for short-lived servers.
- Preserve remote server behavior because remote servers do not receive local Docker container environment variables.
- Add regression coverage for missing required secrets, fallback URI handling, global `--long-lived`, remote boundaries, and short-lived compatibility.

## Risk

This is scoped to required secrets for local long-lived Docker servers. It does not add secret-store watching or container recreation on secret changes, and it does not change short-lived or remote server startup behavior.

## Verification

```bash
go test ./pkg/gateway -run 'TestBuildFallbackURIsSkipsLongLivedLocalSecrets|TestBuildFallbackURIsKeepsShortLivedLocalSecrets|TestBuildFallbackURIsDoesNotBackfillSharedLongLivedSecrets|TestAcquireClientRefusesLongLivedServerWithMissingSecrets|TestAcquireClientRefusesLongLivedServerWithEmptySecretValue|TestAcquireClientRefusesGlobalLongLivedServerWithMissingSecrets|TestAddServerHandlerDoesNotAppendServerWhenValidationFails|TestApplyConfigShortLivedMissingSecretKeepsPlaceholder|TestLongLived' -count=1
go test ./pkg/gateway -count=1
go test ./pkg/gateway ./pkg/catalog ./pkg/workingset -count=1
go test -race ./pkg/gateway -run 'TestAcquireClientRefusesLongLivedServerWithMissingSecrets|TestAcquireClientNoDuplicatesUnderConcurrency|TestBuildFallbackURIsSkipsLongLivedLocalSecrets|TestBuildFallbackURIsDoesNotBackfillSharedLongLivedSecrets' -count=1
go test ./... -short -count=1
```

### DIFF
--- a/pkg/gateway/activateprofile.go
+++ b/pkg/gateway/activateprofile.go
@@ -70,9 +70,10 @@ func (g *Gateway) convertWorkingSetToConfiguration(ctx context.Context, ws worki
 		// Build secrets configs
 		namespace := ""
 		configs = append(configs, ServerSecretConfig{
-			Secrets:   server.Snapshot.Server.Secrets,
-			OAuth:     server.Snapshot.Server.OAuth,
-			Namespace: namespace,
+			Secrets:                server.Snapshot.Server.Secrets,
+			OAuth:                  server.Snapshot.Server.OAuth,
+			Namespace:              namespace,
+			RequireVerifiedSecrets: localLongLivedServer(server.Snapshot.Server, g.LongLived),
 		})
 
 		// Add tools
--- a/pkg/gateway/clientpool.go
+++ b/pkg/gateway/clientpool.go
@@ -83,6 +83,21 @@ func (cp *clientPool) longLived(serverConfig *catalog.ServerConfig, config *clie
 	return serverConfig.Spec.LongLived || cp.LongLived
 }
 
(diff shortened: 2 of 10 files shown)


cutecatfann:
All checks pass and this generally looks good, but misssing some verified signatures on the PR :)

PR #507, opened 2026-06-17
Sec #5: Harden remote MCP URL handling

kgprs:
## Summary

- Adds a centralized remote URL guard that requires public HTTPS destinations by default, rejects userinfo and unsafe hostnames/IP ranges, resolves DNS before use, and validates redirect destinations through a guarded transport.
- Applies the guard to remote MCP client connections and dynamic OAuth discovery/DCR/token HTTP paths that derive from remote MCP URLs.
- Adds regression coverage for loopback, link-local, private, metadata, cluster-local, userinfo, DNS resolution, redirect blocking, NAT64, and the remote client rejection path.

## Developer impact

Local or HTTP remote MCP endpoints now require an explicit dev/test opt-in via `DOCKER_MCP_ALLOW_INSECURE_REMOTE_URLS=1`.

Guarded remote MCP and OAuth HTTP flows intentionally use direct transports instead of Docker Desktop/corporate proxy transports. A forward proxy hides the final target from the client-side dial-time IP validation, so direct dialing is the current security tradeoff for these guarded paths. Environments that require proxy routing may need a follow-up design that can validate the proxy-resolved destination equivalently; this tradeoff should be reviewed with Desktop proxy owners before merge.

## Validation

- `go test ./pkg/remoteurl ./pkg/oauth ./pkg/oauth/dcr ./pkg/oauthdiscovery`
- `go test -short ./pkg/mcp`
- `go test ./cmd/docker-mcp/server -run 'TestEnable'`
- `go test -short ./pkg/...`
- `docker buildx build --no-cache --build-arg GO_LDFLAGS --build-arg DOCKER_MCP_PLUGIN_BINARY --target=lint --platform=darwin .`

Note: `go test -short ./pkg/... ./cmd/docker-mcp/...` still fails in existing command-package tests that attempt to use Docker Desktop proxy sockets / localhost test servers in this environment; the full `pkg/...` short suite and directly affected command-server tests pass.

### DIFF
--- a/pkg/oauth/manager.go
+++ b/pkg/oauth/manager.go
@@ -3,13 +3,11 @@ package oauth
 import (
 	"context"
 	"fmt"
-	"net/http"
 	"net/url"
 
 	"github.com/docker/docker-credential-helpers/credentials"
 	"golang.org/x/oauth2"
 
-	"github.com/docker/mcp-gateway/pkg/desktop"
 	"github.com/docker/mcp-gateway/pkg/log"
 	"github.com/docker/mcp-gateway/pkg/oauth/dcr"
 )
@@ -140,6 +138,9 @@ func (m *Manager) ExchangeCode(ctx context.Context, code string, state string) e
 	if err != nil {
 		return fmt.Errorf("DCR client not found for %s: %w", serverName, err)
 	}
+	if err := validateOutboundDCRClientEndpoints(ctx, dcrClient); err != nil {
+		return err
+	}
 
 	// Create provider
 	provider := NewDCRProvider(dcrClient, m.redirectURI)
@@ -156,10 +157,9 @@ func (m *Manager) ExchangeCode(ctx context.Context, code string, state string) e
 	}
 
 	// Inject proxy transport so the token endpoint is reachable through
-	// Docker Desktop's HTTP proxy when applicable.
-	proxyCtx := context.WithValue(ctx, oauth2.HTTPClient, &http.Client{
-		Transport: desktop.ProxyTransport(),
-	})
+	// Docker Desktop's HTTP proxy when applicable, while blocking unsafe
+	// derived OAuth endpoints.
+	proxyCtx := context.WithValue(ctx, oauth2.HTTPClient, guardedOAuthHTTPClient(0))
 
 	token, err := config.Exchange(proxyCtx, code, opts...)
 	if err != nil {
--- a/pkg/oauthdiscovery/discovery.go
+++ b/pkg/oauthdiscovery/discovery.go
@@ -14,7 +14,6 @@ import (
 
 	oauthhelpers "github.com/docker/mcp-gateway-oauth-helpers"
 
-	"github.com/docker/mcp-gateway/pkg/desktop"
 	"github.com/docker/mcp-gateway/pkg/remoteurl"
 )
 
@@ -23,7 +22,7 @@ func DiscoverOAuthRequirements(ctx context.Context, serverURL string) (*oauthhel
 		return nil, err
 	}
 
-	client := remoteurl.NewHTTPClient(30*time.Second, desktop.ProxyTransport())
+	client := remoteurl.NewDirectHTTPClient(30 * time.Second)
 
 	parsedURL, err := url.Parse(serverURL)
 	if err != nil {
@@ -191,7 +190,7 @@ func PerformDCR(ctx context.Context, discovery *oauthhelpers.Discovery, serverNa
 	req.Header.Set("Accept", "application/json")
 	req.Header.Set("User-Agent", "MCP-Gateway/1.0.0")
 
-	client := remoteurl.NewHTTPClient(30*time.Second, desktop.ProxyTransport())
+	client := remoteurl.NewDirectHTTPClient(30 * time.Second)
 	resp, err := client.Do(req)
 	if err != nil {
 		return nil, fmt.Errorf("failed to send DCR request to %s: %w", discovery.RegistrationEndpoint, err)
@@ -276,7 +275,7 @@ func fetchOAuthProtectedResourceMetadata(ctx context.Context, client *http.Clien
 }
 
 func fetchAuthorizationServerMetadata(ctx context.Context, client *http.Client, authServerURL string) (*oauthhelpers.AuthorizationServerMetadata, error) {
-	metadataURL, err := buildRFC8414WellKnownURL(authServerURL)
+	metadataURL, err := buildRFC8414WellKnownURL(ctx, authServerURL)
 	if err != nil {
 		return nil, fmt.Errorf("building well-known URL: %w", err)
 	}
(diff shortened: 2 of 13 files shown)


pkg/oauth/manager.go
```diff
@@ -156,10 +157,9 @@ func (m *Manager) ExchangeCode(ctx context.Context, code string, state string) e
 	}
 
 	// Inject proxy transport so the token endpoint is reachable through
-	// Docker Desktop's HTTP proxy when applicable.
-	proxyCtx := context.WithValue(ctx, oauth2.HTTPClient, &http.Client{
-		Transport: desktop.ProxyTransport(),
-	})
+	// Docker Desktop's HTTP proxy when applicable, while blocking unsafe
+	// derived OAuth endpoints.
+	proxyCtx := context.WithValue(ctx, oauth2.HTTPClient, guardedOAuthHTTPClient(0))
```
jchangx, line 162: Heads up — this (and the sibling sites) swaps `desktop.ProxyTransport()` for `guardedOAuthHTTPClient(0)`, which uses `GuardDirectTransport()` and sets `Proxy = nil`. That's the correct *security* call (a forward proxy hides the real target behind CONNECT and defeats the dial-time IP pin), but it silently drops **Docker Desktop / corporate proxy support** from every OAuth + remote flow. Anyone behind a proxy whose OAuth previously worked through Desktop's proxy will now attempt direct connections and may fail. Worth (a) calling this out explicitly in the PR description, and (b) confirming with whoever owns the Desktop proxy integration that direct-only is acceptable — it's not mentioned anywhere in the PR today.

pkg/remoteurl/remoteurl.go
```diff
@@ -0,0 +1,338 @@
...
+	netip.MustParsePrefix("198.18.0.0/15"),
+	netip.MustParsePrefix("198.51.100.0/24"),
+	netip.MustParsePrefix("203.0.113.0/24"),
+	netip.MustParsePrefix("224.0.0.0/4"),
+	netip.MustParsePrefix("240.0.0.0/4"),
+	netip.MustParsePrefix("255.255.255.255/32"),
+	netip.MustParsePrefix("::/128"),
+	netip.MustParsePrefix("::1/128"),
+	netip.MustParsePrefix("fc00::/7"),
+	netip.MustParsePrefix("fe80::/10"),
```
jchangx, line 317: Small defense-in-depth gap: the blocklist doesn't include the well-known NAT64 prefix `64:ff9b::/96`. In a NAT64 environment `64:ff9b::<v4>` maps to an embedded IPv4 (incl. metadata/private). One line to close. (The enterprise twin of this guard added it.)

pkg/remoteurl/remoteurl.go
```diff
@@ -0,0 +1,338 @@
...
+		originalDialContext = (&net.Dialer{}).DialContext
+	}
+
+	cloned.DialContext = func(ctx context.Context, network, address string) (net.Conn, error) {
+		host, port, err := net.SplitHostPort(address)
+		if err != nil {
+			return originalDialContext(ctx, network, address)
+		}
+
+		targetHost, _ := ctx.Value(contextKey{}).(string)
```
jchangx: Minor robustness note: the dial guard only validates when `sameHostname(targetHost, host)` and `targetHost` is read from the context value set by `validatingRoundTripper`. If the guarded `*http.Transport` is ever used without that round-tripper wrapper (future refactor, direct use), `targetHost` is empty → `sameHostname` is false → it falls through to `originalDialContext` with no IP validation. Today the wiring always pairs them, so it's fine — but a one-line comment stating that invariant (or asserting the context value is present) would prevent a quiet regression.

jchangx (commented):
Strong PR — solid layered SSRF defense (check-time `Validate` + dial-time IP-pin + `CheckRedirect`), thorough blocked-prefix list, userinfo/zone-ID rejection, IPv4-mapped `Unmap` handled. I traced the real outbound paths (remote client, OAuth discovery/DCR/token exchange) and they all route through guarded clients — no validate-then-plain-fetch holes. No blocking concerns.

The one thing I'd most want surfaced is the **proxy-support removal** (inline below) — it's a real behavior change that isn't in the PR description. The rest are minor/defense-in-depth. Nice work.

pkg/remoteurl/remoteurl.go
```diff
@@ -0,0 +1,314 @@
...
+		if !sameHostname(targetHost, host) || v.allowInsecure {
+			return originalDialContext(ctx, network, address)
```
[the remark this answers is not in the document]
kgprs: Addressed in `33b4beb5`: the guarded remote MCP/OAuth paths now use `GuardDirectTransport` / `NewDirectHTTPClient`, so we do not wrap `desktop.ProxyTransport()` for these guarded outbound calls. That keeps the dial target aligned with the validated request host and avoids the proxy-resolved-target bypass. I also added the proxy-support tradeoff to the PR body under Developer impact.

pkg/oauthdiscovery/discovery.go
```diff
@@ -0,0 +1,365 @@
...
+	return &metadata, nil
+}
+
+func buildRFC8414WellKnownURL(issuerURL string) (string, error) {
+	parsed, err := url.Parse(issuerURL)
+	if err != nil {
+		return "", fmt.Errorf("invalid issuer URL: %w", err)
+	}
+	if parsed.Scheme != "https" {
+		return "", fmt.Errorf("issuer URL must use https scheme")
```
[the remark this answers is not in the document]
kgprs: Addressed in `33b4beb5`: OAuth well-known URL construction now validates the issuer through `remoteurl.Validate(ctx, issuerURL)` and preserves the validated scheme, so `DOCKER_MCP_ALLOW_INSECURE_REMOTE_URLS=1` also covers local HTTP OAuth discovery. Added regression coverage in `pkg/oauthdiscovery/discovery_test.go`.

pkg/oauth/manager.go
kgprs: Called out in the PR body under Developer impact. The direct-only guarded transport is intentional for this PR because a forward proxy hides the final target from client-side dial-time IP validation. I noted that proxy-required environments need a follow-up design that can provide equivalent proxy-resolved destination validation, and that this tradeoff should be reviewed with Desktop proxy owners before merge.

pkg/remoteurl/remoteurl.go
kgprs: Addressed in `3c24eb72`: added `64:ff9b::/96` to the blocked prefix list and added a NAT64 regression case to `TestValidateRejectsUnsafeRemoteURLs`.

pkg/remoteurl/remoteurl.go
kgprs: Addressed in `3c24eb72`: the cloned transport now fails closed if the target host context is missing, instead of silently falling through to the original dialer. I added `TestGuardDialerRequiresRequestTargetHost` to cover that invariant.

jchangx (approved):
Re-reviewed — all my comments are addressed. NAT64 `64:ff9b::/96` added (+test); the dial guard now fail-closes when the request-context target host is missing instead of falling through unvalidated (+test); and the proxy-support removal is now documented in the PR body and acknowledged as an intentional security tradeoff. Nice that the direct-transport choice also resolves the bot's independent proxy-rebinding finding.

Approving. One non-code flag, per your own note in the PR body: the direct-only/no-proxy tradeoff still warrants sign-off from the Docker Desktop proxy owners before merge, since it removes proxy routing from OAuth + remote flows. LGTM otherwise.

PR #508, opened 2026-06-17
Sec #2: Harden container gateway auth

kgprs:
## Summary

- require bearer-token auth for SSE/streaming gateway transports even when running with DOCKER_MCP_IN_CONTAINER=1
- keep container TCP listeners reachable by default for published ports and Compose service-name access, with --host available for explicit bind overrides
- add --allow-unauthenticated as the explicit opt-out and fail closed when HTTP transports start without auth
- add focused tests for container auth setup, listener binding helpers, CLI defaults, and generated CLI reference docs

## Validation

- go test ./pkg/gateway
- go test ./pkg/gateway -run 'Test(InitializeHTTPAuth|GatewayListenAddress|IsExternallyReachableHost)'
- go test ./cmd/docker-mcp/commands -run TestGatewayRunCommandContainerAuthDefaults
- make lint-windows
- make docs

Note: local docker buildx bake validate-docs is blocked in this worktree because Docker sees a .git file pointing at an external worktree path that is not mounted into the build context. The generated docs were refreshed with make docs.

### DIFF
--- a/cmd/docker-mcp/commands/gateway.go
+++ b/cmd/docker-mcp/commands/gateway.go
@@ -39,6 +39,7 @@ func gatewayCommand(docker docker.Client, dockerCli command.Cli, features featur
 				Cpus:         1,
 				Memory:       "2Gb",
 				Transport:    "stdio",
+				Host:         gateway.DefaultContainerGatewayHost,
 				LogCalls:     true,
 				BlockSecrets: true,
 				Verbose:      true,
@@ -203,7 +204,9 @@ func gatewayCommand(docker docker.Client, dockerCli command.Cli, features featur
 	runCmd.Flags().StringArrayVar(&options.OciRef, "oci-ref", options.OciRef, "OCI image references to use")
 	runCmd.Flags().StringSliceVar(&mcpRegistryUrls, "mcp-registry", nil, "MCP registry URLs to fetch servers from (can be repeated)")
 	runCmd.Flags().IntVar(&options.Port, "port", options.Port, "TCP port to listen on (default is to listen on stdio)")
+	runCmd.Flags().StringVar(&options.Host, "host", options.Host, "Host or IP address to bind TCP transports to")
 	runCmd.Flags().StringVar(&options.Transport, "transport", options.Transport, "stdio, sse or streaming. Uses MCP_GATEWAY_AUTH_TOKEN environment variable for localhost authentication to prevent dns rebinding attacks.")
+	runCmd.Flags().BoolVar(&options.AllowUnauthenticated, "allow-unauthenticated", options.AllowUnauthenticated, "Allow unauthenticated HTTP/SSE gateway requests")
 	runCmd.Flags().BoolVar(&options.LogCalls, "log-calls", options.LogCalls, "Log calls to the tools")
 	runCmd.Flags().BoolVar(&options.BlockSecrets, "block-secrets", options.BlockSecrets, "Block secrets from being/received sent to/from tools")
 	runCmd.Flags().BoolVar(&options.BlockNetwork, "block-network", options.BlockNetwork, "Block tools from accessing forbidden network resources")
--- a/cmd/docker-mcp/commands/gateway.go
+++ b/cmd/docker-mcp/commands/gateway.go
@@ -39,7 +39,6 @@ func gatewayCommand(docker docker.Client, dockerCli command.Cli, features featur
 				Cpus:         1,
 				Memory:       "2Gb",
 				Transport:    "stdio",
-				Host:         gateway.DefaultContainerGatewayHost,
 				LogCalls:     true,
 				BlockSecrets: true,
 				Verbose:      true,
--- a/pkg/gateway/run.go
+++ b/pkg/gateway/run.go
@@ -492,8 +492,11 @@ func (g *Gateway) initializeHTTPAuth() error {
 		return nil
 	}
 	if g.AllowUnauthenticated {
+		if os.Getenv("MCP_GATEWAY_AUTH_TOKEN") != "" {
+			log.Log("Warning: ignoring MCP_GATEWAY_AUTH_TOKEN because --allow-unauthenticated is set")
+		}
 		if isExternallyReachableHost(g.Host) {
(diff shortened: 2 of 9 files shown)


pkg/gateway/run.go
```diff
@@ -226,7 +227,7 @@ func (g *Gateway) Run(ctx context.Context) error {
 			lc  net.ListenConfig
 			err error
 		)
-		ln, err = lc.Listen(ctx, "tcp", fmt.Sprintf(":%d", port))
+		ln, err = lc.Listen(ctx, "tcp", gatewayListenAddress(g.Host, port))
```
jchangx, line 230: Not a regression (matches `main`), but worth flagging in a hardening PR: with `--host` unset, `gatewayListenAddress("", port)` → `":port"`, which binds **all interfaces (0.0.0.0)**, not loopback. Mandatory auth mostly mitigates that — but combined with `--allow-unauthenticated` you get an open, unauthenticated tool-execution endpoint reachable on every interface, surfaced only by a single log line. Consider defaulting the non-container case to `127.0.0.1` (require explicit `--host 0.0.0.0` for published-port/Compose use), or at minimum making the `--allow-unauthenticated` + external-bind combination warn much more loudly.

pkg/gateway/run.go
```diff
@@ -492,6 +487,60 @@ func (g *Gateway) Run(ctx context.Context) error {
 	}
 }
 
+func (g *Gateway) initializeHTTPAuth() error {
+	if !isHTTPTransport(g.Transport) {
+		return nil
+	}
+	if g.AllowUnauthenticated {
```
jchangx, line 494: Footgun worth a log line: when `--allow-unauthenticated` is set this returns before `getOrGenerateAuthToken`, so a present `MCP_GATEWAY_AUTH_TOKEN` is silently discarded and the endpoint runs open. Defensible precedence (explicit flag beats ambient env), but an operator who set a token expecting enforcement gets an open endpoint with only a generic warning. Either log specifically "ignoring MCP_GATEWAY_AUTH_TOKEN because --allow-unauthenticated is set," or flip precedence so a present token forces auth on.

pkg/gateway/transport.go
```diff
@@ -19,6 +20,10 @@ func (g *Gateway) startStdioServer(ctx context.Context, _ io.Reader, _ io.Writer
 }
 
 func (g *Gateway) startSseServer(ctx context.Context, ln net.Listener) error {
+	if g.authToken == "" && !g.AllowUnauthenticated {
```
jchangx, line 23: This fail-closed guard is the most important behavior in the PR, but I don't think it's covered by a test — the existing tests assert the building blocks (token sourcing, listen-address, allow-unauth skip) but none asserts that `startSseServer`/`startStreamingServer` *errors* with `authToken==""` + `AllowUnauthenticated==false`. Worth adding a direct test so the headline "can't start open unless you opt out" invariant is locked in.

jchangx (commented):
The approach is correct and the headline goal is achieved — the old "silently disable auth when `DOCKER_MCP_IN_CONTAINER=1`" footgun is genuinely fixed, with fail-closed gates so the only way to start an HTTP/SSE/streaming transport without auth is an explicit `--allow-unauthenticated`. No blocking concerns.

One thing for the description/release notes: this is a **breaking change** for existing container deployments that ran SSE/streaming auth-free — they'll now fail to start unless they set `MCP_GATEWAY_AUTH_TOKEN` or pass `--allow-unauthenticated`. That's the safe (fail-closed) direction, but it needs a migration callout so deployments aren't unexpectedly locked out. Inline notes below are all minor.

cmd/docker-mcp/commands/gateway.go
```diff
@@ -39,6 +39,7 @@ func gatewayCommand(docker docker.Client, dockerCli command.Cli, features featur
 				Cpus:         1,
 				Memory:       "2Gb",
 				Transport:    "stdio",
+				Host:         gateway.DefaultContainerGatewayHost,
```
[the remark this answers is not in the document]
kgprs: Addressed by keeping the default listener reachable instead of forcing container loopback. The current code leaves an unset `--host` as `:port`/all interfaces, so published-port and Compose flows stay reachable without requiring every container invocation to pass `--host 0.0.0.0` explicitly. This thread is outdated against the earlier default.

pkg/gateway/run.go
kgprs: Handled in `afe219b2`. I kept the unset-host behavior for compatibility with existing published-port/Compose use, but made the `--allow-unauthenticated` + externally reachable listener case much louder: it now logs an uppercase `WARNING`, names the listener (`all interfaces` for an empty host), calls out that reachable clients can execute configured MCP tools, and suggests removing `--allow-unauthenticated` or binding to `127.0.0.1`. The test now asserts that warning content for the all-interfaces case.

pkg/gateway/run.go
kgprs: Handled in `afe219b2`. Explicit `--allow-unauthenticated` still wins over the ambient environment token, but the gateway now logs `Warning: ignoring MCP_GATEWAY_AUTH_TOKEN because --allow-unauthenticated is set` before running open, so that precedence is visible to operators.

pkg/gateway/transport.go
kgprs: Handled in `afe219b2`. Added `TestHTTPTransportsRequireAuthToken`, which directly exercises both `startSseServer` and `startStreamingServer` with an empty auth token and default `AllowUnauthenticated == false`, and asserts that both fail closed before trying to serve. Local `go test ./pkg/gateway`, `make lint-windows`, and the PR checks are green on the pushed commit.

jchangx (approved):
Re-reviewed — all three points addressed: the `--allow-unauthenticated` + present-token case now logs an explicit "ignoring MCP_GATEWAY_AUTH_TOKEN" warning, the all-interfaces exposure warning is now loud and actionable (names the bind + the tool-execution risk + the remediation), and `TestHTTPTransportsRequireAuthToken` now locks in the fail-closed behavior for SSE/streaming. Approving. (Still worth a release-note/migration callout that auth-free container deployments will now fail closed — safe direction, just needs to be surfaced to operators.)

PR #509, opened 2026-06-17
Sec #7: Constrain local file inputs

kgprs:
## Summary
- Route `file://` server-definition reads and legacy catalog reads through the trusted Docker MCP catalog-directory resolver before opening files.
- Constrain local catalog names and local file references to paths that resolve under `~/.docker/mcp/catalogs`, including symlink-resolved containment checks.
- Return the validated real path from the resolver to avoid validating one path and opening another.
- Decode escaped `file://` path components before applying the trusted-root boundary, while documenting Docker MCP's catalog-root-relative `file://` behavior.
- Update CLI help/generated docs and add regression coverage for traversal, symlink escape, untrusted absolute/dot-relative paths, resolved-path returns, and escaped local file references.

## Breaking change
- Local catalog paths supplied via `--catalog` or `--additional-catalog` must now resolve under `~/.docker/mcp/catalogs`. Absolute or `./` catalog paths outside that directory are rejected instead of being read directly.
- `file://` server-definition paths are likewise constrained to the same trusted catalog directory.

## Validation
- `go test ./pkg/catalog ./pkg/workingset ./pkg/catalog_next ./pkg/gateway`
- `go test ./pkg -run '^$'`
- `go test ./cmd/docker-mcp/commands -run '^$'`
- `make lint-darwin`
- `make docs`

### DIFF
--- a/cmd/docker-mcp/commands/gateway.go
+++ b/cmd/docker-mcp/commands/gateway.go
@@ -189,8 +189,8 @@ func gatewayCommand(docker docker.Client, dockerCli command.Cli, features featur
 		runCmd.Flags().StringVar(&options.WorkingSet, "profile", "", "Profile ID to use (mutually exclusive with --servers and --enable-all-servers)")
 	}
 	runCmd.Flags().BoolVar(&enableAllServers, "enable-all-servers", false, "Enable all servers in the catalog (instead of using individual --servers options)")
-	runCmd.Flags().StringSliceVar(&options.CatalogPath, "catalog", options.CatalogPath, "Paths to docker catalogs (absolute or relative to ~/.docker/mcp/catalogs/)")
-	runCmd.Flags().StringSliceVar(&additionalCatalogs, "additional-catalog", nil, "Additional catalog paths to append to the default catalogs")
+	runCmd.Flags().StringSliceVar(&options.CatalogPath, "catalog", options.CatalogPath, "Paths to docker catalogs under ~/.docker/mcp/catalogs/")
+	runCmd.Flags().StringSliceVar(&additionalCatalogs, "additional-catalog", nil, "Additional catalog paths under ~/.docker/mcp/catalogs/ to append to the default catalogs")
 	runCmd.Flags().StringSliceVar(&options.RegistryPath, "registry", options.RegistryPath, "Paths to the registry files (absolute or relative to ~/.docker/mcp/)")
 	runCmd.Flags().StringSliceVar(&additionalRegistries, "additional-registry", nil, "Additional registry paths to merge with the default registry.yaml")
 	runCmd.Flags().StringSliceVar(&options.ConfigPath, "config", options.ConfigPath, "Paths to the config files (absolute or relative to ~/.docker/mcp/)")
--- a/cmd/docker-mcp/commands/gateway.go
+++ b/cmd/docker-mcp/commands/gateway.go
@@ -189,8 +189,8 @@ func gatewayCommand(docker docker.Client, dockerCli command.Cli, features featur
 		runCmd.Flags().StringVar(&options.WorkingSet, "profile", "", "Profile ID to use (mutually exclusive with --servers and --enable-all-servers)")
 	}
 	runCmd.Flags().BoolVar(&enableAllServers, "enable-all-servers", false, "Enable all servers in the catalog (instead of using individual --servers options)")
-	runCmd.Flags().StringSliceVar(&options.CatalogPath, "catalog", options.CatalogPath, "Paths to docker catalogs under ~/.docker/mcp/catalogs/")
-	runCmd.Flags().StringSliceVar(&additionalCatalogs, "additional-catalog", nil, "Additional catalog paths under ~/.docker/mcp/catalogs/ to append to the default catalogs")
+	runCmd.Flags().StringSliceVar(&options.CatalogPath, "catalog", options.CatalogPath, "Catalog paths must resolve under ~/.docker/mcp/catalogs/")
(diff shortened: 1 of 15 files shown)


pkg/catalog/catalog.go
```diff
@@ -128,6 +119,67 @@ func readFileOrURL(ctx context.Context, fileOrURL string) ([]byte, error) {
...
+	candidateForCheck, err := evalPathIfExists(candidate)
+	if err != nil {
+		return "", err
+	}
+
+	if !pathWithin(root, candidateForCheck) {
+		return "", errUntrustedLocalPath
+	}
+
+	return candidate, nil
```
jchangx: TOCTOU: this validates `candidateForCheck` (the `EvalSymlinks`-resolved path) but returns the **un-resolved** `candidate`, which callers then `os.ReadFile` — re-walking and re-following symlinks at open time. That reopens a check-to-use window: anything able to write inside the catalogs dir (or a parent component) can swap a component for a symlink after validation passes, so the bytes actually read can come from outside the boundary. Cheap fix: return `candidateForCheck` so the read targets the already-validated real path (and it's more self-consistent — today you validate one path and hand back a different one). FWIW the enterprise sibling of this change (#1949) already does it this way.

cmd/docker-mcp/commands/gateway.go
```diff
@@ -189,8 +189,8 @@ func gatewayCommand(docker docker.Client, dockerCli command.Cli, features featur
 		runCmd.Flags().StringVar(&options.WorkingSet, "profile", "", "Profile ID to use (mutually exclusive with --servers and --enable-all-servers)")
 	}
 	runCmd.Flags().BoolVar(&enableAllServers, "enable-all-servers", false, "Enable all servers in the catalog (instead of using individual --servers options)")
-	runCmd.Flags().StringSliceVar(&options.CatalogPath, "catalog", options.CatalogPath, "Paths to docker catalogs (absolute or relative to ~/.docker/mcp/catalogs/)")
-	runCmd.Flags().StringSliceVar(&additionalCatalogs, "additional-catalog", nil, "Additional catalog paths to append to the default catalogs")
+	runCmd.Flags().StringSliceVar(&options.CatalogPath, "catalog", options.CatalogPath, "Paths to docker catalogs under ~/.docker/mcp/catalogs/")
```
jchangx: Two things to reconcile here. (1) The PR description says it "preserves explicit absolute and ./ catalog paths" — but the code doesn't keep them as an *escape*: absolute/`./` paths still must pass containment, so an absolute path outside the catalogs dir is rejected. That's good security, but it contradicts the description. (2) This new help ("under ~/.docker/mcp/catalogs/") now disagrees with `--registry`/`--config`, which still accept absolute paths and don't go through this resolver — so an operator who kept catalogs at an absolute path alongside registries/configs will silently get "no catalog" or an untrusted-path error. Please reconcile description + help, and call out the removed absolute-path support as a breaking change.

jchangx (commented):
Sound and well-executed — one resolver funnels all catalog reads, contains to `~/.docker/mcp/catalogs`, and (importantly) checks containment on the **symlink-resolved** path with a component-aware `filepath.Rel` check, which correctly avoids the `/catalogsEVIL`-matches-`/catalogs` prefix bug. The symlink-escape test exercises a real symlink, not just lexical `..`. No blocking concerns; a couple of items worth addressing — two inline below, plus one here.

**`file://` parsing (`pkg/workingset/workingset.go`, `ResolveServersFromString`, ~L398):** `file://` is split with `strings.CutPrefix`, not parsed as a URL, so the mapping is inconsistent — `file://host/path` becomes a relative name `host/path` (silently wrong) and percent-encoding is never decoded. It's safe today only because everything still funnels through `ResolveLocalCatalogPath` and containment catches escapes, but relying on "the resolver catches it anyway" is fragile. Suggest `url.Parse`, reject a non-empty `Host`, and use the decoded `Path`; or at minimum a comment that `file://` is deliberately treated as a raw path with containment as the only defense. (Couldn't anchor this inline — the line is unchanged in the diff.)

pkg/catalog/catalog.go
```diff
@@ -110,13 +113,11 @@ func readFileOrURL(ctx context.Context, fileOrURL string) ([]byte, error) {
 		return buf, nil
 
 	default:
-		homeDir, err := user.HomeDir()
+		path, err := ResolveLocalCatalogPath(fileOrURL)
```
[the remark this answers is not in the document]
kgprs: Fixed in `c826b384`: all non-URL catalog reads now go through `ResolveLocalCatalogPath`, including absolute and `./` inputs, so local catalog reads are constrained consistently. I also added regression coverage for absolute, dot-relative, and traversal inputs outside `~/.docker/mcp/catalogs`.

kgprs:
Thanks for the review. I addressed the unanchored `file://` parsing note in `bae8b04b`: local file references now decode percent escapes before applying the trusted-root resolver, and the code documents that Docker MCP treats `file://` as a catalog-root-relative local file reference rather than an RFC file URI with a host component. I kept that existing `file://testdata/server.yaml` / `file://./server.yaml` behavior intentional instead of rejecting non-empty URI hosts, because current CLI examples and tests depend on catalog-relative file references.

jchangx (approved):
Re-reviewed — all addressed. The TOCTOU is fixed (`ResolveLocalCatalogPath` now returns the EvalSymlinks-resolved path, +test), the help text/description is reconciled across the catalog flags + regenerated docs, and `file://` now goes through `parseLocalFileReference`, which documents the catalog-root-relative intent and — nicely — `url.PathUnescape`s before the containment check, so encoded traversal like `%2e%2e` canonicalizes to `..` and is rejected. Approving.

PR #510, opened 2026-06-17
Sec #1: Harden Docker bind mounts

kgprs:
## Summary
- Validate Docker volume specs before adding `-v` args for server and tool containers.
- Require host bind mounts to be read-only and under trusted roots.
- Resolve symlinks and expand `~/` with the shared home-dir helper to avoid allowlist bypasses.

## Impact
Unsafe host-path binds now fail before Docker is invoked. Named volumes and read-only binds under temp or `MCP_GATEWAY_DOCKER_BIND_ALLOWED_PATHS` continue to work.

## Validation
- `go test -count=1 ./pkg/gateway`
- `go test -count=1 ./pkg/gateway -run 'TestValidateDockerVolumeBinds|TestApplyConfigMountAs|TestApplyConfigLongLivedRejectsWritableHostBind|TestApplyConfigShortLivedRejectsWritableHostBind|TestArgsAndEnv_SkipsFlagShapedValues'`\n\nNote: `go test -count=1 ./pkg/...` was also attempted, but existing Docker-backed tests in `pkg/gateway/embeddings` and `pkg/mcp` fail in this environment because Docker credential helper `docker-credential-osxkeychain` is unavailable.

### DIFF
--- a/pkg/gateway/docker_binds.go
+++ b/pkg/gateway/docker_binds.go
@@ -20,6 +20,8 @@ const dockerBindAllowedPathsEnv = "MCP_GATEWAY_DOCKER_BIND_ALLOWED_PATHS"
 type dockerVolumeBind struct {
 	raw        string
 	source     string
+	target     string
+	mode       string
 	hostPath   bool
 	readOnly   bool
 	sourcePath string
@@ -28,25 +30,36 @@ type dockerVolumeBind struct {
 func validateDockerVolumeBinds(volumes []string) error {
 	allowedRoots := dockerBindAllowedRoots()
 	for _, raw := range volumes {
-		bind, err := parseDockerVolumeBind(raw)
-		if err != nil {
+		if _, err := normalizeDockerVolumeBindWithRoots(raw, allowedRoots); err != nil {
 			return err
-		}
-		if !bind.hostPath {
-			continue
-		}
-		if !bind.readOnly {
-			return fmt.Errorf("unsafe docker volume %q: host path bind mounts must be read-only", bind.raw)
-		}
-		if disallowed, reason := disallowedDockerHostPath(bind.sourcePath); disallowed {
-			return fmt.Errorf("unsafe docker volume %q: host path %q is blocked (%s)", bind.raw, bind.source, reason)
-		}
-		if !isPathUnderAnyRoot(bind.sourcePath, allowedRoots) {
-			return fmt.Errorf("unsafe docker volume %q: host path %q is outside allowed roots %s",
-				bind.raw, bind.source, strings.Join(allowedRoots, ", "))
 		}
 	}
 	return nil
+}
+
+func normalizeDockerVolumeBind(raw string) (string, error) {
+	return normalizeDockerVolumeBindWithRoots(raw, dockerBindAllowedRoots())
+}
+
+func normalizeDockerVolumeBindWithRoots(raw string, allowedRoots []string) (string, error) {
+	bind, err := parseDockerVolumeBind(raw)
+	if err != nil {
+		return "", err
+	}
+	if !bind.hostPath {
+		return bind.raw, nil
+	}
+	if !bind.readOnly {
+		return "", fmt.Errorf("unsafe docker volume %q: host path bind mounts must be read-only", bind.raw)
+	}
+	if disallowed, reason := disallowedDockerHostPath(bind.sourcePath); disallowed {
+		return "", fmt.Errorf("unsafe docker volume %q: host path %q is blocked (%s)", bind.raw, bind.source, reason)
+	}
+	if !isPathUnderAnyRoot(bind.sourcePath, allowedRoots) {
+		return "", fmt.Errorf("unsafe docker volume %q: host path %q is outside allowed roots %s",
+			bind.raw, bind.source, strings.Join(allowedRoots, ", "))
+	}
+	return bind.sourcePath + ":" + bind.target + ":" + bind.mode, nil
 }
 
 func parseDockerVolumeBind(raw string) (dockerVolumeBind, error) {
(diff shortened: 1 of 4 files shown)


pkg/gateway/docker_binds.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	if strings.HasPrefix(p, "~/") {
+		home, err := user.HomeDir()
+		if err != nil {
+			return "", fmt.Errorf("expand home directory: %w", err)
+		}
+		if home == "" {
+			return "", fmt.Errorf("expand home directory: empty home directory")
+		}
+		p = path.Join(filepath.ToSlash(home), strings.TrimPrefix(p, "~/"))
```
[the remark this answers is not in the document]
kgprs: Addressed in 72d2b5c5. The Docker volume validation path now returns a normalized bind spec, and both Docker call sites append that normalized value instead of the original raw mount. That means `~/trusted/project:/data:ro` is expanded before it reaches `exec.Command`/`docker run -v`. Added coverage asserting the normalized output for the home-root case, and the PR checks are green.

pkg/gateway/docker_binds.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+	return readOnly
+}
+
+func dockerVolumeSourceIsHostPath(source string) bool {
+	source = strings.TrimSpace(source)
+	if source == "" {
+		return false
+	}
+	return strings.HasPrefix(source, "/") ||
```
jchangx: **Blocking — validation bypass.** This classifies a source as a host path only when it starts with `/`, `\\`, `./`, `../`, a drive letter, or `~/`. A *relative* source containing an embedded slash — e.g. `subdir/../../../../etc:/data` or even `foo/bar:/data` — matches none of these, so it's treated as a **named volume** and skips all validation. It's then passed to `docker run -v` (clientpool.go), and Docker interprets any `-v` source containing `/` as a **bind path relative to CWD**, so `..` segments escape upward (to /etc, the socket dir, etc.). `isSafeFlagValue` only rejects a leading `-`, so it doesn't catch this.

Fix: invert the test like the enterprise twin (#1945) — treat a source as a host path whenever it is **not** a syntactically valid Docker volume name (i.e. it contains `/`, `\`, `.`, `~`, or a drive colon), rather than enumerating positive prefixes. At minimum, also classify any source containing `/` or `..` as a host path. Please add `subdir/../../../etc:/data` and `foo/bar:/data` as reject-test cases — they currently pass through unvalidated.

pkg/gateway/docker_binds.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+
+	resolvedExisting, err := filepath.EvalSymlinks(existing)
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(append([]string{resolvedExisting}, missing...)...), nil
+}
+
+func dockerBindAllowedRoots() []string {
```
jchangx, line 253: Non-blocking, but a posture call worth surfacing: with no env var set, this still allows host binds of `/tmp`, `/private/tmp`, `/var/tmp`, and `os.TempDir()` (RO). So a malicious spec can RO-mount any victim/attacker-writable temp path (another process's temp secrets, predictable temp files) with zero operator opt-in. The enterprise twin is fail-closed (no env → no host binds). If the temp-dir allowance is intentional for local dev, please call it out in the PR/docs; otherwise consider matching enterprise and failing closed.

jchangx (commented):
Good hardening and the test breadth is solid (socket reject, symlink-escape-to-/etc, credential paths, RO enforcement). But I found **one blocking validation bypass** (inline on `dockerVolumeSourceIsHostPath`) — a relative source containing `/` or `..` is misclassified as a *named volume* and skips all validation, then is handed to `docker run -v` which treats it as a CWD-relative bind that `..` can escape upward from. That's the exact class this PR exists to prevent, so I can't approve until it's fixed.

Worth noting: the enterprise twin (#1945) is **not** vulnerable to this because it uses the inverse rule — anything that isn't a syntactically valid Docker volume name is treated as a host path and validated. Bringing this PR in line with that approach fixes the bypass.

Two non-blocking notes inline + one here: `embeddings/client.go:44` constructs a writable `-v %s:/data` bind that doesn't route through the new validator. `dataDir` is gateway-controlled (not a server-spec value), so it's outside this PR's malicious-spec threat model — but it's the one remaining unguarded `-v` site, so a follow-up to route it through `normalizeDockerVolumeBind` (or a comment on why it's exempt) would make the "all binds go through one validator" invariant actually hold.

pkg/gateway/docker_binds.go
kgprs: Called out the default explicitly in the bind-policy comment: the no-env default is a read-only temp-root allowlist. Keeping that behavior intentional for local/dev `mount_as` use while still requiring RO, symlink resolution, sensitive-path blocking, and credential-path blocking. If maintainers prefer enterprise-style fail-closed behavior here too, that can be a follow-up policy change.

pkg/gateway/docker_binds.go
kgprs: Fixed in 7f0b982b. `dockerVolumeSourceIsHostPath` now treats a source as a named volume only when it matches the documented Docker volume-name shape (`[a-zA-Z0-9][a-zA-Z0-9_.-]*`); everything else is treated as a host path and must pass the bind policy. Added reject coverage for the exact examples `foo/bar:/data` and `subdir/../../../etc:/data`, plus read-only variants that verify they are rejected as outside allowed roots.

kgprs:
Addressed the latest review feedback in 7f0b982b:

- Fixed the relative-source validation bypass by treating only syntactically valid Docker volume names as named volumes; all other sources go through host-path validation.
- Added regression tests for `foo/bar:/data` and `subdir/../../../etc:/data`, plus read-only variants that exercise allowed-root rejection.
- Clarified the default read-only temp-root allowance in the bind-policy comment.
- Added a comment at `pkg/gateway/embeddings/client.go` explaining why the remaining direct writable `-v` is exempt: `dataDir` is gateway-selected state storage and the vector DB needs it writable.

Local validation: `go test -count=1 ./pkg/gateway`, `go test -run "^$" ./pkg/gateway/embeddings`, and `git diff --check`.

jchangx (approved):
Re-reviewed — the blocking bypass is fixed correctly. `dockerVolumeSourceIsHostPath` now returns `!dockerVolumeSourceIsNamedVolume(source)`, with named-volume membership gated on the Docker volume-name charset (alnum start, then `[A-Za-z0-9_.-]`) — so any source containing `/`, `\`, `~`, etc. is treated as a host path and validated. My two bypass cases (`foo/bar:/data`, `subdir/../../../etc:/data`) are now reject-tests, and the embeddings writable mount is documented as gateway-selected/exempt. This matches the enterprise twin (#1945). Approving.

One residual nit (non-blocking, your call): the default still allows RO host binds of the temp roots (`/tmp`, `$TMPDIR`, …) with no opt-in — now documented as the intended OSS default. It's low-risk (read-only, temp) and reasonable for local dev, but enterprise is fail-closed; if you ever want parity, failing closed here too would remove the implicit temp-dir read surface. Not blocking.

PR #511, opened 2026-06-17
Sec #5: Allow guarded Docker Desktop proxy sockets

kgprs:
## Summary

- Adds an explicit Docker Desktop proxy socket transport/dialer helper that ignores arbitrary HTTP proxy environment variables.
- Adds a `remoteurl` trusted-proxy dialer guard that still validates request and redirect URLs, while permitting the known Docker Desktop local proxy socket to dial the final target.
- Wires guarded remote MCP, OAuth discovery/DCR, and OAuth token exchange clients to use the Desktop proxy socket when Desktop is available, with direct guarded transport as the fallback.
- Falls back to direct transport when the Desktop proxy socket is unavailable: Unix requires an actual socket file, and Windows probes the Docker Desktop named pipe with a bounded dial before selecting the proxy transport.
- Keeps Docker-image integration tests opt-in where they depend on unavailable external images, and only preserves `PATH` for stdio MCP server commands so Docker credential helpers remain discoverable without inheriting the full gateway environment.

## Why

The remote URL hardening in #507 intentionally switched guarded remote/OAuth flows to direct transports because generic forward proxies hide the final target from client-side dial-time IP validation. Docker Desktop’s local proxy socket is a trusted local integration point, so this follow-up restores Desktop proxy socket support explicitly without re-enabling arbitrary corporate/environment proxy routing for these guarded paths.

The trusted Desktop proxy path still validates request and redirect URLs before the proxy is dialed. Unlike direct transport, it cannot pin the post-validation dialed IP because Docker Desktop’s proxy resolves the final target internally; the code now documents that accepted residual DNS-rebinding race and has a negative test ensuring blocked DNS results reject before the proxy dialer is invoked.

## Validation

- go test -count=1 ./...
- go test -short ./pkg/... ./cmd/docker-mcp/...
- GOOS=windows GOARCH=amd64 go test -c ./pkg/desktop -o /tmp/mcp-gateway-pkg-desktop-windows.test.exe
- docker buildx build --build-arg GO_LDFLAGS --build-arg DOCKER_MCP_PLUGIN_BINARY --target=test .
- make lint

### DIFF
--- a/pkg/desktop/raw_client.go
+++ b/pkg/desktop/raw_client.go
@@ -37,23 +37,34 @@ func ProxyTransport() http.RoundTripper {
 			return
 		}
 
-		ctx := context.Background()
-		if !IsRunningInDockerDesktop(ctx) {
+		transport := DockerDesktopProxySocketTransport(context.Background())
+		if transport == nil {
 			desktopProxyTransportInst = http.DefaultTransport
 			return
 		}
-
-		desktopProxyTransportInst = &http.Transport{
-			Proxy: http.ProxyURL(&url.URL{
-				Scheme: "http",
-			}),
-			DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
-				return dialHTTPProxy(ctx)
-			},
-		}
+		desktopProxyTransportInst = transport
 	})
 
 	return desktopProxyTransportInst
+}
+
+// DockerDesktopProxySocketTransport returns a transport that proxies through
+// Docker Desktop's local HTTP proxy socket only. It intentionally ignores
+// HTTP_PROXY/HTTPS_PROXY environment variables so callers can opt into the
+// trusted Desktop socket without permitting arbitrary forward proxies.
+func DockerDesktopProxySocketTransport(ctx context.Context) http.RoundTripper {
+	if !IsRunningInDockerDesktop(ctx) {
+		return nil
+	}
+
+	return &http.Transport{
+		Proxy: http.ProxyURL(&url.URL{
+			Scheme: "http",
+		}),
+		DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
+			return dialHTTPProxy(ctx)
+		},
+	}
 }
 
 // hasEnvProxyVars returns true if any HTTP proxy environment variables are set.
--- a/pkg/desktop/raw_client.go
+++ b/pkg/desktop/raw_client.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"runtime"
 	"sync"
 	"time"
 )
@@ -56,6 +57,9 @@ func DockerDesktopProxySocketTransport(ctx context.Context) http.RoundTripper {
 	if !IsRunningInDockerDesktop(ctx) {
 		return nil
 	}
+	if !desktopProxySocketAvailable() {
+		return nil
+	}
 
 	return &http.Transport{
 		Proxy: http.ProxyURL(&url.URL{
@@ -65,6 +69,18 @@ func DockerDesktopProxySocketTransport(ctx context.Context) http.RoundTripper {
 			return dialHTTPProxy(ctx)
 		},
 	}
+}
+
+func desktopProxySocketAvailable() bool {
+	if runtime.GOOS == "windows" {
+		return true
+	}
+
+	info, err := os.Stat(Paths().HTTPProxySocket)
+	if err != nil {
+		return false
+	}
+	return info.Mode().Type() == os.ModeSocket
 }
 
 // hasEnvProxyVars returns true if any HTTP proxy environment variables are set.
--- a/pkg/desktop/raw_client.go
+++ b/pkg/desktop/raw_client.go
@@ -10,7 +10,6 @@ import (
 	"net/http"
 	"net/url"
 	"os"
-	"runtime"
 	"sync"
 	"time"
 )
@@ -57,7 +56,7 @@ func DockerDesktopProxySocketTransport(ctx context.Context) http.RoundTripper {
 	if !IsRunningInDockerDesktop(ctx) {
 		return nil
 	}
-	if !desktopProxySocketAvailable() {
+	if !desktopProxySocketAvailable(ctx) {
 		return nil
 	}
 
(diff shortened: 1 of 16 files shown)


pkg/desktop/raw_client.go
```diff
@@ -37,25 +38,51 @@ func ProxyTransport() http.RoundTripper {
...
+		}),
+		DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
+			return dialHTTPProxy(ctx)
+		},
+	}
+}
+
+func desktopProxySocketAvailable() bool {
+	if runtime.GOOS == "windows" {
+		return true
```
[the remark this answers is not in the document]
kgprs: Addressed in `4e9deda7`; follow-up lint cleanup is in `55191584`.

Windows now probes the Docker Desktop HTTP proxy named pipe with a bounded dial before returning the trusted proxy transport. If the pipe is absent or unavailable, `DockerDesktopProxySocketTransport` returns nil and callers use the direct guarded fallback. Non-Windows still requires the path to be an actual Unix socket.

I added coverage for the Windows missing-pipe and successful-probe paths, plus non-Windows missing/regular-file/socket cases.

Validation run on the latest head:
- `go test -count=1 ./...`
- `go test -short ./pkg/... ./cmd/docker-mcp/...`
- `GOOS=windows GOARCH=amd64 go test -c ./pkg/desktop -o /tmp/mcp-gateway-pkg-desktop-windows.test.exe`
- `docker buildx build --build-arg GO_LDFLAGS --build-arg DOCKER_MCP_PLUGIN_BINARY --target=test .`
- `make lint`

GitHub checks are passing on `55191584`.

pkg/remoteurl/remoteurl.go
```diff
@@ -201,9 +232,15 @@ func (v Validator) guardDialer(base http.RoundTripper) http.RoundTripper {
 			return nil, fmt.Errorf("remote URL target host missing from request context")
 		}
 
-		if !sameHostname(targetHost, host) || v.allowInsecure {
+		if v.allowInsecure {
 			return originalDialContext(ctx, network, address)
 		}
+		if !sameHostname(targetHost, host) {
+			if allowTrustedProxy {
```
jchangx, line 254: Residual rebinding gap worth acknowledging: for the direct transport, `guardDialer` re-resolves at dial time and pins to a validated IP, which catches TOCTOU DNS-rebinding. Under `allowTrustedProxy`, the backend name is resolved *inside* Docker Desktop's proxy and never re-validated — so the only check on the real target is the single `ValidateURL` DNS lookup in `RoundTrip`. An attacker controlling a public hostname with a very short TTL could pass that lookup (public IP) and have the proxy re-resolve to 127.0.0.1/169.254.x a moment later. Much narrower than the hole #507 closed (requires winning a race against the proxy's resolver, only under Desktop), but real. Ideally pin the validated IP into the proxied connection (dial the validated literal / `CONNECT <ip>`); at minimum document it as accepted residual risk in this function's comment.

pkg/remoteurl/remoteurl.go
```diff
@@ -132,6 +132,13 @@ func GuardTransport(base http.RoundTripper) http.RoundTripper {
 	return DefaultValidator().GuardTransport(base)
 }
 
+// GuardTrustedProxyTransport validates request and redirect URLs, but permits
+// dial targets that differ from the request host. Use it only with a trusted
+// local proxy transport such as Docker Desktop's proxy socket.
+func GuardTrustedProxyTransport(base http.RoundTripper) http.RoundTripper {
```
jchangx: Minor API foot-gun: the dial-pin relaxation is keyed purely on the `allowTrustedProxy` bool, not on any property of the proxy — its safety depends entirely on callers only ever passing the hardcoded Desktop-socket transport. Today they do, so it's safe, but a future caller passing an env-derived/attacker-influenced proxy transport would silently get the pin disabled. Consider taking the Desktop socket dialer directly (rather than an arbitrary `http.RoundTripper`), or asserting the transport's dialer is the Desktop one, so this can't be misused.

pkg/mcp/stdio.go
```diff
@@ -63,6 +63,13 @@ func (c *stdioMCPClient) Initialize(ctx context.Context, _ *mcp.InitializeParams
 	return nil
 }
 
+func commandEnv(env []string) []string {
+	if len(env) == 0 {
+		return os.Environ()
```
jchangx: Unrelated to the proxy work and worth its own justification: `commandEnv` now merges `os.Environ()` into every stdio MCP server's child-process env, so stdio servers inherit the gateway's full process environment (including any secrets in env). Likely intended for credential-helper PATH discovery, but it's an env-exposure surface change bundled into a security PR — ideally split out or called out explicitly with the rationale.

jchangx (dismissed):
Reviewed this carefully since it re-opens the surface #507 closed. **Verdict: safe to approve.** The SSRF defense is two-layered and this relaxes only Layer 2 (the dial-time IP pin), which legitimately cannot apply when traffic egresses through a proxy. Layer 1 — `ValidateURL` resolving the real request host and rejecting blocked IP ranges (loopback/link-local/private/metadata) — still runs on every request and every redirect for all re-enabled flows (remote MCP, OAuth token/discovery/DCR), *before* the proxy is ever dialed. The proxy endpoint is hardcoded to the Docker Desktop socket (gated on IsRunningInDockerDesktop + socket existence, and it explicitly ignores HTTP(S)_PROXY), so it is not attacker-influenceable. Nice — this restores proxy support (the regression I flagged on #507) without reopening the hole.

Nothing blocking. Inline notes are non-blocking hardening + one unrelated-change flag. One test recommendation: add a negative case — trusted-proxy transport + a fake resolver mapping a host to `169.254.169.254`/`127.0.0.1` → assert the request is rejected and the dialer is NOT called. That pins the security-critical property (Layer 1 still gates the proxy path) against future regressions; the current new tests only exercise `public.example.test → 8.8.8.8`.

jchangx (approved):
Re-reviewed the latest round — all three of my notes are addressed, and well:

- **API foot-gun (was the main one): structurally fixed.** `GuardTrustedProxyTransport(base http.RoundTripper)` is replaced by `GuardTrustedProxyDialer(dialer TrustedProxyDialer)`, so the dial-pin relaxation is now tied to *providing a trusted local socket dialer* rather than keyed on a bool over an arbitrary transport. A caller can no longer pass an env-derived/attacker-influenced proxy transport and silently disable the pin. 
- **Residual proxy-side rebinding: documented as accepted** in the `GuardTrustedProxyDialer` doc comment (validator still rejects blocked request/redirect URLs before the proxy is dialed; the narrow rebinding race is explicitly accepted to preserve Desktop proxy semantics). Good — that was exactly the ask.
- **Negative test added:** `TestGuardTrustedProxyDialerRejectsUnsafeBeforeProxyDial` maps a host to `169.254.169.254` and asserts the request is rejected ("resolved to disallowed address") with the dialer never called — pins the Layer-1-still-gates-the-proxy property against regressions.

And a bonus improvement: `commandEnv` was tightened from merging the full `os.Environ()` into stdio child processes down to injecting only `PATH` (when the server didn't set it) — that removes the broad env/secret-leakage surface I flagged. Nice.

Nits only now — re-affirming approval.

PR #514, opened 2026-06-18
Sec #5: Harden untrusted URL fetches

kgprs:
## Summary

- Route untrusted catalog, README, and MCP registry fetches through `pkg/remoteurl` guarded direct HTTP clients.
- Validate parsed MCP registry URLs with the remote URL guard before deriving registry API endpoints.
- Add regression coverage for unsafe loopback/default HTTP rejection and explicit insecure test opt-in.

## Why

Follow-up to #507: several non-MCP-client fetch paths still used `desktop.ProxyTransport()` directly, which bypassed the host/range validation model added for remote MCP URLs.

## Validation

- `go test -short ./pkg/...`
- `go test ./pkg/fetch ./pkg/catalog ./pkg/catalog_next ./pkg/registryapi ./pkg/workingset`
- `go test ./cmd/docker-mcp/commands -run 'TestMcpregistryImportCommand'`\n\nNote: the full `go test ./cmd/docker-mcp/commands` still hits existing Docker Desktop proxy socket assumptions in `TestDockerCatalogProtection` in this environment.

### DIFF
--- a/cmd/docker-mcp/commands/import.go
+++ b/cmd/docker-mcp/commands/import.go
@@ -5,22 +5,16 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"net/url"
+	"time"
 
 	"github.com/docker/mcp-gateway/pkg/catalog"
-	"github.com/docker/mcp-gateway/pkg/desktop"
 	"github.com/docker/mcp-gateway/pkg/oci"
+	"github.com/docker/mcp-gateway/pkg/remoteurl"
 )
 
 func runMcpregistryImport(ctx context.Context, serverURL string, servers *[]catalog.Server) error {
-	// Validate URL
-	parsedURL, err := url.Parse(serverURL)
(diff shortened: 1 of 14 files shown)


bobbyhouse:
Non-blocking question: `catalog.readFileOrURL` and `registryapi.NewClient` now use `DirectTransport()` which sets `Proxy = nil`, bypassing any `HTTP_PROXY`/`HTTPS_PROXY` environment variables. This is the same tradeoff already accepted for MCP/OAuth paths in #507, but it now extends to catalog and registry API fetches.

Worth a quick check: will this be a problem for BYOC or hosted customer deployments where outbound HTTPS traffic to `desktop.docker.com` or `registry.modelcontextprotocol.io` routes through a corporate proxy? If those environments rely on proxy env vars for those fetches today, they would silently break after this change.

bobbyhouse (approved):

jchangx (approved):

PR #515, opened 2026-06-18
Sec #3: Verify MCP images before pulling

kgprs:
## Summary
- Verify Docker MCP `mcp/...` image signatures before pulling server images.
- Route dynamic server add/profile activation pulls through the same pull-and-verify helper.
- Enable signature verification by default for `docker mcp gateway run` and update command docs.

## Why
The gateway had signature verification infrastructure, but runtime image pulls could still happen without enforcement unless verification was explicitly enabled. Dynamic server add/profile activation paths also called `PullImage` directly. This makes Docker MCP images fail closed by default while leaving non-`mcp/...` third-party images outside Docker-key verification.

## Validation
- `go test ./pkg/gateway`
- `go test ./cmd/docker-mcp/commands -run 'TestGateway'`\n- `go test ./pkg/gateway ./pkg/docker ./pkg/signatures`\n- `git diff --check`\n\nNote: `go test ./cmd/docker-mcp/commands ./pkg/gateway` fails in existing command package tests (`TestDockerCatalogProtection`, `TestMcpregistryImportCommand`) because the test environment points HTTP traffic at an invalid Docker Desktop proxy socket under the test temp home.

### DIFF
--- a/pkg/gateway/pull.go
+++ b/pkg/gateway/pull.go
@@ -67,13 +67,18 @@ func (g *Gateway) pullImages(ctx context.Context, images []string) error {
 }
 
 func (g *Gateway) verifyImages(ctx context.Context, images []string) error {
-	if !g.VerifySignatures || len(images) == 0 {
+	if len(images) == 0 {
+		return nil
+	}
+
+	if !g.VerifySignatures {
+		log.Log("Warning: signature verification is disabled; MCP server images will not be verified and may use mutable tags")
 		return nil
 	}
 
 	for _, image := range images {
 		if !strings.Contains(image, "@sha256:") {
-			return fmt.Errorf("verifying docker image %s: image must be referenced by digest", image)
+			return fmt.Errorf("verifying docker image %s: image must be referenced by digest; pin the MCP image to a sha256 digest or disable signature verification with --verify-signatures=false", image)
 		}
 	}
 
--- a/pkg/gateway/pull.go
+++ b/pkg/gateway/pull.go
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/distribution/reference"
+
 	"github.com/docker/mcp-gateway/pkg/log"
 	"github.com/docker/mcp-gateway/pkg/signatures"
 )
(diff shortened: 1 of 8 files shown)


pkg/gateway/pull.go
```diff
@@ -49,21 +67,41 @@ func (g *Gateway) pullImages(ctx context.Context, images []string) error {
 }
 
 func (g *Gateway) verifyImages(ctx context.Context, images []string) error {
-	if !g.VerifySignatures {
+	if !g.VerifySignatures || len(images) == 0 {
```
jchangx: By-design escape hatch, but worth a louder signal: with `--verify-signatures=false`, MCP images are pulled and run with no verification *and* no digest requirement — the one remaining path to run unverified. Since it's the only off-switch, consider logging a visible warning when verification is explicitly disabled so it can't be flipped off unnoticed in a deployment.

pkg/gateway/pull.go
```diff
@@ -49,21 +67,41 @@ func (g *Gateway) pullImages(ctx context.Context, images []string) error {
...
 		return nil
 	}
 
+	for _, image := range images {
+		if !strings.Contains(image, "@sha256:") {
+			return fmt.Errorf("verifying docker image %s: image must be referenced by digest", image)
```
jchangx: Correct fail-closed posture, but it's a behavior change worth a release note: a custom `--catalog` that lists an MCP server by tag (`mcp/foo:latest`) now hard-fails at startup with this error, where it worked on main (verify was best-effort + tag-tolerant). The official catalog pins digests so production is fine; just make the upgrade path explicit (pin a digest, or `--verify-signatures=false`). A one-line hint in the error pointing at those two options would help.

jchangx (dismissed):
Reviewed carefully — solid, well-scoped security fix. Approving.

The core win: it fixes the verify-before-pull ordering (main pulled *then* verified, with no digest requirement), and TOCTOU is now genuinely closed for MCP images — `verifyImages` rejects any `mcp/...` ref not pinned by `@sha256:`, the cosign verifier verifies that digest, and the container runs the same string with `--pull never`, so verify-digest == pull-digest == run-digest. `isDockerMCPImage` is also correctly anchored (parses with `ParseNormalizedNamed` and matches `docker.io`/`index.docker.io` + `mcp/` path), so `ghcr.io/mcp/evil` / `evilmcp/...` can't masquerade. Default-on reaches the CLI and the container/Helm entrypoint with no hidden off-switch. cosign + Rekor against a pinned embedded key, fail-closed on both missing and bad signatures.

No blocking issues — a couple of minor/by-design notes inline, plus two test gaps worth filling (non-blocking): there's no test asserting a *failed* verification blocks the pull (the verifier stub always returns nil), and no negative test that a spoofed `mcp/` host (`ghcr.io/mcp/...`) is NOT treated as MCP — that host-anchoring is the security-critical bit and isn't directly asserted.

jchangx (approved):

PR #519, opened 2026-06-23
Validate profile config before gateway startup

kgprs:
## Summary
- share profile config validation between dynamic profile activation and profile-backed gateway startup
- fail profile startup with a missing/invalid config error before evaluating Docker volume mounts
- add regression coverage for a Kubernetes profile missing `config_path`

## Why This Is Needed
`docker mcp gateway run --profile <id>` uses `WorkingSetConfiguration` to load profile-backed servers at startup. That path was not running the same required config validation that `mcp-activate-profile` runs for dynamically activated profiles.

In v0.43.0, Docker volume handling was hardened to reject malformed bind mounts before invoking Docker. That is correct, but it exposed missing profile config as a low-level Docker volume error. For the Kubernetes catalog entry, the volume template is:

```yaml
volumes:
  - "{{kubernetes.config_path}}:/home/appuser/.kube/config"
config:
  - name: kubernetes
    required:
      - config_path
```

When `config_path` is missing from the profile, the template evaluates to:

```text
:/home/appuser/.kube/config
```

The gateway then reports:

```text
Can't start kubernetes: validate volume for kubernetes: invalid docker volume ":/home/appuser/.kube/config": source and target are required
```

That message is technically true, but it points users at Docker volume syntax instead of the real problem: the profile is missing required Kubernetes config.

With this change, profile startup fails before container arg construction with the same style of validation error used by profile activation, for example:

```text
Cannot activate profile 'test'. Validation failed for 1 server(s):

Server 'kubernetes':
  Missing/invalid config: config_path (missing)
```

A user can then fix the profile directly:

```sh
docker mcp profile config test --set kubernetes.config_path=$HOME/.kube/config
```

## Validation
- `go test ./pkg/gateway`
- `go test ./pkg/gateway ./pkg/workingset`

### DIFF
--- a/pkg/gateway/activateprofile.go
+++ b/pkg/gateway/activateprofile.go
@@ -9,7 +9,6 @@ import (
 	"slices"
 	"strings"
 
-	"github.com/google/jsonschema-go/jsonschema"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 
 	"github.com/docker/mcp-gateway/pkg/catalog"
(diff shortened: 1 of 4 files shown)


pkg/gateway/config_validation.go
```diff
@@ -0,0 +1,166 @@
...
+				if requiredProps[propName] {
+					missingConfig = append(missingConfig, fmt.Sprintf("%s (missing)", propName))
+				}
+				continue
+			}
+			if isEmptyConfigValue(configValue) {
+				if requiredProps[propName] {
+					missingConfig = append(missingConfig, fmt.Sprintf("%s (missing)", propName))
+				}
+				continue
```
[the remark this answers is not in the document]
kgprs: Addressed in `e8bbf861`. `isEmptyConfigValue` now only treats `nil` and empty strings as missing, so an empty object like `headers: {}` is passed through to JSON Schema validation. I also added `TestValidateServerConfigAllowsEmptyObjectValue` to cover the required empty-object case.

pkg/gateway/activateprofile.go
```diff
@@ -143,81 +142,7 @@ func (g *Gateway) ActivateProfile(ctx context.Context, ws workingset.WorkingSet)
...
-					if err := resolved.Validate(configValue); err != nil {
-						errMsg := err.Error()
-						if len(errMsg) > 100 {
-							errMsg = errMsg[:97] + "..."
-						}
-						validation.missingConfig = append(validation.missingConfig, fmt.Sprintf("%s (%s)", propName, errMsg))
-					}
-				}
-			}
+			validation.missingConfig = append(validation.missingConfig, validateServerConfig(serverConfig, profileConfig.config[serverName])...)
```
saucow, line 145: `ActivateProfile` is the other caller of the shared `validateServerConfig`, and this refactor moved its inline validation + error formatting out into the new helpers (`validateServerConfig` + `formatProfileValidationError`). But there's no test exercising this path — only the startup/`readOnce` path is covered (`TestWorkingSetConfigurationReadRejectsMissingRequiredConfig`); a `grep` for `ActivateProfile` across `*_test.go` finds nothing.

Worth adding a test that calls `ActivateProfile` with a server missing a required config value (and ideally also a missing secret) and asserts the aggregated message — `Cannot activate profile '…'` / `Server '…'` / `Missing/invalid config: … (missing)` — so the wiring (does the handler still collect `missingConfig` and aggregate it with `missingSecrets`/`imagePullError`?) is locked, not just the helper in isolation.

pkg/gateway/config_validation.go
```diff
@@ -0,0 +1,163 @@
...
+			}
```
saucow, line 82: `isEmptyConfigValue` returns before `resolved.Validate`, so the schema-validation branch below (the `resolved.Validate(configValue)` block) is the *only* thing keeping the empty-`{}` relaxation narrow: a required string-typed field set to `{}` or `[]` is caught only there, not by `isEmptyConfigValue`. That backstop currently has no test — a one-line future change broadening `isEmptyConfigValue` to also treat empty maps/slices as "skip" would silently drop the type-mismatch check with nothing failing.

A negative test would pin it: `validateServerConfig` should reject `config_path: {}` (and `config_path: []`) for a string-typed required field. (`TestValidateServerConfigAllowsEmptyObjectValue` covers the *allow* side; this covers the *reject* side that makes the allow safe.)

Minor, separate: an empty *optional* string now skips `minLength`/`pattern` validation it previously hit. Almost certainly fine — the registry import schema generator doesn't emit those constraints — just flagging the behavior change for hand-authored catalog/profile schemas.

saucow (commented):
**Review — validate profile config before startup**

Nice change. The key win is net-new: `readOnce` (the profile-backed startup path) had *no* config validation on `main`, and this adds it fail-closed at `configuration_workingset.go:77-78`, before any volume eval — so the cryptic Docker bind error (`invalid docker volume ":/home/appuser/.kube/config"`) is replaced by the same actionable message activation already produced. The extraction into `validateServerConfig` reads faithful to the deleted inline logic, and the two behavioral changes both *tighten* (the new `[]string` `required` arm, which the old `[]any`-only cast silently skipped; and rejecting present-but-empty required values). It doesn't touch the bind or URL hardening seams.

**Recommendation: approve-with-nits** — two test-coverage gaps and one optional edge case, all inline; none blocking.

saucow (dismissed):
Approving — this adds net-new fail-closed config validation on the profile-backed startup path (which had none), the extraction is faithful to the deleted inline logic, and both behavioral changes tighten. The items in my earlier review are test-coverage gaps (the untested `ActivateProfile` path and the schema-reject backstop) plus one optional edge case — none are blocking.

pkg/gateway/activateprofile.go
kgprs: Addressed in `60153709`. Added `TestActivateProfileRejectsMissingRequiredConfig`, which exercises `ActivateProfile` with a remote server snapshot, missing required config, and a missing secret against an empty secrets-engine response. It asserts the aggregated `Cannot activate profile`, `Server`, `Missing secrets`, and `Missing/invalid config` message pieces.

pkg/gateway/config_validation.go
kgprs: Addressed in `60153709`. Added `TestValidateServerConfigRejectsWrongTypeForRequiredString` covering `config_path: {}` and `config_path: []` against a required string schema, so those values are rejected by JSON Schema validation rather than treated as missing or skipped.

saucow (approved):
Re-reviewed — the test additions resolve both of my earlier comments:

- `TestActivateProfileRejectsMissingRequiredConfig` drives the `ActivateProfile` → `validateServerConfig` wiring end to end and asserts the aggregated error (including the missing-secret case).
- `TestValidateServerConfigRejectsWrongTypeForRequiredString` locks the schema-reject backstop for `config_path: {}` / `[]`, so the empty-object relaxation stays narrow.

Verified the package tests pass locally. Approving — thanks for the quick turnaround.

PR #523, opened 2026-06-23
use policies for server load and config set

mickael-docker:
mirrors the behaviour of `mcp-add` and provides defense-in-depth

### DIFF
--- a/pkg/gateway/activateprofile.go
+++ b/pkg/gateway/activateprofile.go
@@ -131,6 +131,9 @@ func (g *Gateway) ActivateProfile(ctx context.Context, ws workingset.WorkingSet)
 	var validationErrors []serverValidation
 
 	for _, serverName := range serversToActivate {
+		if err := g.checkServerLoadPolicy(ctx, serverName, nil); err != nil {
+			return err
+		}
 		serverConfig := profileConfig.servers[serverName]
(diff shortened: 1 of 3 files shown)


pkg/gateway/activateprofile.go
```diff
@@ -131,6 +131,9 @@ func (g *Gateway) ActivateProfile(ctx context.Context, ws workingset.WorkingSet)
 	var validationErrors []serverValidation
 
 	for _, serverName := range serversToActivate {
+		if err := g.checkServerLoadPolicy(ctx, serverName, nil); err != nil {
```
kgprs: Because `checkServerLoadPolicy` always builds the request from `g.configuration`, this path evaluates profile servers before they have been merged into that configuration. For a newly activated profile server, `Configuration.policyRequest` hits the missing-server branch and sends only the server/action, without the target, server type, server source, transport, catalog, or working-set metadata that policy rules may depend on. That means policies matching those fields can fail to block `activate-profile`. Consider evaluating against `profileConfig` here, or passing an already-built `policy.Request` into the shared helper.

kgprs (commented):
I found one issue in the activate-profile policy path.

kgprs (approved):

PR #524, opened 2026-06-23
update startup warning for auth and local bind

mickael-docker:
**What I did**
Updates the message from #508, ensuring it is presented in all cases


kgprs (approved):
Reviewed diff, focused tests, and CI; no findings.

PR #529, opened 2026-06-30
refactor: use docker/secrets-engine SDK and writable bind mounts in Dockerfile

joe0BAB:
## What I did

- **Use the `docker/secrets-engine` SDK in `secretsengine.go`**: Replaced the hand-rolled connectRPC-over-Unix-socket implementation in the secret-management package with the public `github.com/docker/secrets-engine` client SDK.
  - `GetSecrets`/`GetSecret` now use `client.New` + `Resolver.GetSecrets` instead of manually constructing HTTP requests, JSON payloads, and the RPC path.
  - Use `realms.DockerMCPDefault` for the `docker/mcp/**` pattern and `client.ParsePattern` for single-key lookups.
  - Pin the client to the engine socket via `api.DefaultSocketPath()`, dropping the local `socketPath()`/`newHTTPClient()` helpers.
  - Return the SDK's `client.Envelope` directly (ID is now `secrets.ID`); callers that used the ID as a string now call `.ID.String()`.
  - Update the config-validation test fake to serve the real resolver connect protocol so an empty result maps to `ErrSecretNotFound`.

- **Replace rsync with writable bind mounts in the Dockerfile**: Replaced the tmpfs + rsync pattern in the `docs-build` and `docs-validate` stages with a writable bind mount (`--mount=type=bind,target=.,rw`). Avoids copying the full source tree and removes the rsync dependency from the base image. Build output is still exported via `/out` and scratch stages.

- **Bump Go version**: Bumped the `go` directive to 1.25.11 (required by the SDK modules) and the Dockerfile `GO_VERSION` to match. Vendored accordingly.


go.mod
```diff
@@ -187,28 +191,32 @@ require (
...
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20260128011058-8636f8732409 // indirect
+	google.golang.org/grpc v1.78.0 // indirect
+	google.golang.org/protobuf v1.36.11 // indirect
 	gotest.tools/v3 v3.5.2 // indirect
 	k8s.io/client-go v0.33.1 // indirect
 )
 
 replace github.com/modelcontextprotocol/registry => github.com/docker/mcp-community-registry v0.0.0-20251024214917-21000e320421
+
+replace github.com/docker/secrets-engine/client => /Users/joe/work/secrets-engine/client
```
kgprs: This should not be committed with local filesystem replaces. Outside of this machine, any non-vendor module load fails before compiling the SDK imports; for example, `go list -mod=mod -deps ./cmd/docker-mcp/secret-management/secret` reports that `/Users/joe/work/secrets-engine/...` does not exist. Please remove these local replaces and depend on a real published tag/pseudo-version for the SDK modules.

kgprs (changes requested):
Found one blocking issue in the module metadata.

kgprs (dismissed):

go.mod
joe0BAB: removed

kgprs (approved):

PR #534, opened 2026-07-15
refactor(secret): thread secrets-engine client.ID through credstore

joe0BAB:
Rework the secret credstore to use the docker/secrets-engine SDK types throughout instead of raw strings:

- Derive namespaced keys from the canonical realm patterns (realms.DockerMCP*) via Pattern.ExpandID, replacing the hand-rolled namespace prefix constants.
- GetDefaultSecretKey / GetOAuthKey / GetDCRKey now take and return client.ID and return an error (no panic, no expandRealm helper).
- All Set/Delete credstore funcs and List now traffic in client.ID; List skips entries that are not valid secret IDs.
- GetSecret accepts a client.ID rather than a string key.
- Remove ValidateSecretName; callers validate names via client.ParseID directly, pushing parsing to the outer boundary.
- Propagate client.ID end-to-end through the OAuth credential helper (TokenExists, GetTokenStatus and their per-mode handlers) and all callers (gateway run/mcpadd, oauth ls/revoke/cleanup, remote mcp).

**What I did**

**Related issue**

### DIFF
--- a/cmd/docker-mcp/commands/secret.go
+++ b/cmd/docker-mcp/commands/secret.go
@@ -4,9 +4,9 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"slices"
 	"strings"
 
+	seclient "github.com/docker/secrets-engine/client"
 	"github.com/spf13/cobra"
 
 	"github.com/docker/mcp-gateway/cmd/docker-mcp/secret-management/formatting"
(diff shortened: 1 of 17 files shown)


cmd/docker-mcp/commands/secret.go
```diff
@@ -42,18 +42,27 @@ func rmSecretCommand() *cobra.Command {
...
+			var errs []error
+			var ids []seclient.ID
 			if all {
-				var err error
-				ids, err = secret.List(cmd.Context())
+				listed, err := secret.List(cmd.Context())
 				if err != nil {
 					return err
 				}
+				ids = listed
```
kgprs: secret.List returns the fully qualified IDs emitted by docker pass ls (for example, docker/mcp/foo), but DeleteDefaultSecret treats its argument as a relative name and expands it again. In the --all path this therefore asks docker pass to remove docker/mcp/docker/mcp/foo instead of the listed key (and OAuth/DCR IDs are likewise re-prefixed). Please either delete listed IDs directly here or make List return IDs relative to the default realm, and add coverage for rm --all.

cmd/docker-mcp/commands/secret.go
```diff
@@ -42,18 +42,24 @@ func rmSecretCommand() *cobra.Command {
...
 				if err != nil {
 					return err
 				}
-			}
-
-			var errs []error
-			for _, s := range ids {
-				errs = append(errs, secret.DeleteDefaultSecret(cmd.Context(), s))
+				for _, id := range listed {
+					errs = append(errs, secret.DeleteByID(cmd.Context(), id))
```
kgprs, line 53: secret.List calls docker pass ls without a realm filter and its contract says it returns all keychain IDs. Deleting every listed ID verbatim makes docker mcp secret rm --all erase unrelated Docker Pass secrets such as docker/auth/..., docker/sandbox/..., or arbitrary myapp/... keys, not just docker/mcp/**. Please restrict this loop to IDs matching realms.DockerMCPDefault (or expose a realm-scoped list) before DeleteByID, and cover a non-MCP ID in the regression test.

joe0BAB: rewrote to make things more explicit

kgprs (approved):
Re-reviewed the updated head. The --all path now deletes listed IDs verbatim only after filtering to the docker/mcp/** realm, with regression coverage for unrelated Docker Pass keys. No further findings.

PR #546, opened 2026-08-11
Harden OAuth lifecycle and tool policy enforcement

kgprs:
## Summary

This PR closes a group of security gaps across the OAuth lifecycle, remote MCP credential forwarding, and tool invocation policy. The main behavioral change is that revoking an OAuth connection now attempts RFC 7009 revocation at the authorization server before deleting the local credential, so a locally removed token is not left active at the provider.

It also binds credentials and OAuth metadata to the identity that was originally configured: remote HTTP headers cannot cross an origin boundary, credential-helper keys cannot alias through path-like provider names, and authorization-server metadata must identify the exact issuer that was requested.

## Why

The previous behavior had several independent weaknesses:

- OAuth revoke removed the local credential without invalidating refresh and access tokens at providers that support revocation.
- Configured headers, including bearer credentials, could follow a redirect to a different scheme, host, or port.
- OAuth credential keys were built by concatenating unescaped endpoint and provider strings, allowing different identities to map to the same legacy key.
- Authorization-server metadata was accepted without verifying that its `issuer` exactly matched the requested authorization server.
- POCI/container tools did not consistently pass through invoke policy, audit emission, and the standard tool telemetry path.
- The dynamic-server registry could be mistaken for an authorization allowlist even though it represents the active working set.

### Provider-aware OAuth revocation

- Discover and retain `revocation_endpoint` metadata through DCR and token refresh.
- Revoke the refresh token first and then a distinct access token, including `token_type_hint` and the public `client_id`.
- Accept any successful 2xx provider response.
- Never follow redirects while sending bearer tokens.
- Preserve the local token and DCR metadata when provider revocation fails so the operation can be retried.
- When the provider does not advertise revocation support, emit an explicit note and perform local cleanup.
- If no community DCR registration exists, treat revoke as idempotent: skip provider calls, attempt token/DCR and stale Desktop cleanup best-effort, and return success. Other DCR lookup failures remain errors.

### OAuth identity and credential storage

- Require the authorization-server metadata `issuer` to exactly match the requested issuer.
- Validate all discovered OAuth endpoints before use.
- Store tokens under a versioned key whose authorization-endpoint and provider components are independently base64url encoded.
- Transparently migrate unambiguous legacy entries.
- Never read or delete an ambiguous legacy key for provider names containing `/`.

### Remote MCP credential forwarding

- Bind configured and OAuth-derived headers to the original URL origin.
- Reject cross-origin requests before attaching credentials.
- Allow same-origin redirects, including equivalent default ports.
- Reject redirects that change the scheme, host, or effective port.

### Tool policy and observability

- Apply `invoke` policy evaluation to POCI/container tool calls.
- Emit the standard audit event for allowed, denied, and failed policy evaluations.
- Record the existing tool-call counter, duration, error metric, and trace status for every POCI path, including policy errors/denials, argument decoding failures, and container results.
- Document that the dynamic registry is an active set, not an authorization boundary. Deployments that require a fixed allowlist should use `--servers`; deployments using dynamic tools should enforce catalog activation with policy.

### 1. Build the branch

```console
make docker-mcp
./dist/docker-mcp --help
```

Expected: the plugin builds successfully and the standalone help command exits successfully.

### 2. Verify successful provider revocation

Prerequisite: configure `<server>` against an OAuth test provider that advertises an RFC 7009 `revocation_endpoint`, issues distinct refresh and access tokens, and records incoming requests.

```console
./dist/docker-mcp oauth authorize <server> --open-browser
./dist/docker-mcp oauth revoke <server>
```

Expected:

- the provider records a POST for the refresh token with `token_type_hint=refresh_token`;
- it then records a POST for the distinct access token with `token_type_hint=access_token`;
- both requests contain the registered public `client_id`;
- no redirect is followed;
- the command prints `OAuth access revoked for <server>`;
- the next use of the server requires authorization again.

### 3. Verify fail-closed cleanup and retry

1. Configure the test provider's revocation endpoint to return HTTP 500.
2. Authorize `<server>`.
3. Run:

```console
./dist/docker-mcp oauth revoke <server>
```

Expected: the command exits non-zero with a provider-revocation error, and the existing local token/DCR entry remains usable for a retry.

4. Change the endpoint to return HTTP 204 and rerun the same command.

Expected: provider revocation succeeds and local OAuth state is removed.

### 4. Verify a provider without revocation support

1. Remove `revocation_endpoint` from the provider metadata.
2. Authorize and revoke `<server>`.

Expected: the command prints a note that the provider does not advertise revocation, deletes the local token, and completes successfully.

### 5. Verify idempotent revoke before authorization

1. Choose a community server that has never been authorized and confirm it has no local token or DCR entry.
2. Run:

```console
./dist/docker-mcp oauth revoke <never-authorized-server>
```

Expected: no provider endpoint is called; missing local token/DCR deletions are reported only as notes; stale Desktop cleanup is attempted; the command prints `OAuth access revoked for <never-authorized-server>` and exits successfully. A non-not-found DCR lookup failure should still exit non-zero.

### 6. Verify issuer binding

1. Configure the requested authorization server as `https://auth.example.test/issuer-a`.
2. Return metadata whose `issuer` is `https://auth.example.test/issuer-b`.
3. Start authorization.

Expected: discovery fails with an issuer-mismatch error before DCR or token exchange. Restoring the exact requested issuer allows authorization to proceed.

### 7. Verify redirect credential isolation

1. Configure a remote MCP test server at origin A with an `Authorization` or custom secret header.
2. Have origin A redirect the MCP request to origin B, where B differs by scheme, host, or effective port.
3. Run the gateway:

```console
./dist/docker-mcp gateway run --servers <server> --verbose
```

4. Invoke any tool from `<server>`.

Expected: the request is rejected as a different-origin redirect and origin B receives no configured or OAuth header. A redirect to another path on origin A should still work and should retain the header.

### 8. Verify POCI invoke policy

1. Run a catalog server exposing a POCI/container tool with an invoke policy that denies that tool.
2. Invoke the tool through the gateway.

Expected: the container is not executed, the client receives a policy-denied error, an audit decision is emitted, the tool error metric/span status is recorded, and `mcp.tool.duration` receives exactly one sample for the denied call.

3. Send a raw call for the same POCI tool with malformed JSON arguments.

Expected: argument decoding fails before container execution; the error metric/span status is recorded; `mcp.tool.duration` still receives exactly one sample. Change the policy to allow and retry with valid arguments; the tool should execute and record normal counter/duration telemetry.

## Automated validation

- `make test`
- `make lint`
- focused OAuth revocation, discovery, token-store migration, remote transport, gateway policy, audit, and telemetry tests
- all required GitHub checks are green

### DIFF
--- a/cmd/docker-mcp/oauth/revoke.go
+++ b/cmd/docker-mcp/oauth/revoke.go
@@ -23,6 +23,9 @@ var (
 	// requiring docker pass, Docker Desktop, or a real database.
 	deleteOAuthTokenFunc      = secret.DeleteOAuthToken
 	deleteDCRClientFunc       = secret.DeleteDCRClient
+	getCommunityDCRClientFunc = pkgoauth.GetDCRClientFromDockerPass
+	getCommunityTokenFunc     = pkgoauth.GetTokenFromDockerPass
+	revokeProviderTokenFunc   = pkgoauth.RevokeTokenAtProvider
 	desktopDeleteOAuthAppFunc = func(ctx context.Context, app string) error {
 		return desktop.NewAuthClient().DeleteOAuthApp(ctx, app)
 	}
@@ -84,10 +87,10 @@ func revokeCEMode(ctx context.Context, app string) error {
 	credHelper := pkgoauth.NewReadWriteCredentialHelper()
 	manager := pkgoauth.NewManager(credHelper)
 
-	// Delete OAuth token
+	// Revoke at the provider before deleting local credentials. If remote
+	// revocation fails, preserve both the token and DCR metadata for retry.
 	if err := manager.RevokeToken(ctx, app); err != nil {
-		// Token might not exist, continue to DCR deletion
-		fmt.Printf("Note: %v\n", err)
+		return fmt.Errorf("failed to revoke OAuth access: %w", err)
 	}
 
 	// Delete DCR client (matches Desktop behavior)
@@ -109,10 +112,21 @@ func revokeCommunityMode(ctx context.Context, app string) error {
 		return fmt.Errorf("invalid server name %q: %w", app, err)
 	}
 
-	// Delete OAuth token from docker pass
+	dcrClient, err := getCommunityDCRClientFunc(ctx, app)
+	if err != nil {
+		return fmt.Errorf("failed to load DCR client for provider revocation: %w", err)
+	}
+	token, err := getCommunityTokenFunc(ctx, app)
+	if err != nil {
+		return fmt.Errorf("failed to load OAuth token for provider revocation: %w", err)
+	}
+	if err := revokeProviderTokenFunc(ctx, dcrClient, token); err != nil {
+		return fmt.Errorf("failed to revoke OAuth access: %w", err)
+	}
+
+	// Delete OAuth token from docker pass only after provider revocation.
 	if err := deleteOAuthTokenFunc(ctx, appID); err != nil {
-		// Token might not exist, continue to DCR deletion
-		fmt.Printf("Note: %v\n", err)
+		return fmt.Errorf("provider token revoked but failed to delete local OAuth token: %w", err)
 	}
 
 	// Delete DCR client from docker pass (soft failure -- entry may not exist
--- a/cmd/docker-mcp/oauth/revoke.go
+++ b/cmd/docker-mcp/oauth/revoke.go
@@ -2,6 +2,7 @@ package oauth
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	seclient "github.com/docker/secrets-engine/client"
@@ -121,6 +122,18 @@ func revokeCommunityMode(ctx context.Context, app string) error {
 
 	dcrClient, err := getCommunityDCRClientFunc(ctx, app)
 	if err != nil {
+		if errors.Is(err, secret.ErrSecretNotFound) {
+			fmt.Printf("Note: %v\n", err)
+			if err := deleteOAuthTokenFunc(ctx, appID); err != nil {
+				fmt.Printf("Note: %v\n", err)
+			}
+			if err := deleteDCRClientFunc(ctx, appID); err != nil {
+				fmt.Printf("Note: %v\n", err)
+			}
+			cleanStaleDesktopEntriesFunc(ctx, app)
+			fmt.Printf("OAuth access revoked for %s\n", app)
+			return nil
+		}
 		return fmt.Errorf("failed to load DCR client for provider revocation: %w", err)
 	}
 	providerRevoked := false
--- a/cmd/docker-mcp/oauth/revoke.go
+++ b/cmd/docker-mcp/oauth/revoke.go
@@ -13,6 +13,11 @@ import (
 	"github.com/docker/mcp-gateway/pkg/workingset"
 )
 
+type ceRevokeManager interface {
+	RevokeToken(context.Context, string) error
+	DeleteDCRClient(string) error
+}
+
 // Function pointers for testability (same pattern as pkg/workingset/oauth.go).
 var (
 	revokeCEModeFunc        = revokeCEMode
@@ -26,6 +31,9 @@ var (
 	getCommunityDCRClientFunc = pkgoauth.GetDCRClientFromDockerPass
 	getCommunityTokenFunc     = pkgoauth.GetTokenFromDockerPass
 	revokeProviderTokenFunc   = pkgoauth.RevokeTokenAtProvider
+	newCERevokeManagerFunc    = func() ceRevokeManager {
+		return pkgoauth.NewManager(pkgoauth.NewReadWriteCredentialHelper())
+	}
 	desktopDeleteOAuthAppFunc = func(ctx context.Context, app string) error {
 		return desktop.NewAuthClient().DeleteOAuthApp(ctx, app)
 	}
@@ -84,8 +92,7 @@ func revokeDesktopMode(ctx context.Context, app string) error {
 // revokeCEMode handles revoke in standalone CE mode
 // Matches Desktop behavior: deletes both token and DCR client
 func revokeCEMode(ctx context.Context, app string) error {
-	credHelper := pkgoauth.NewReadWriteCredentialHelper()
-	manager := pkgoauth.NewManager(credHelper)
+	manager := newCERevokeManagerFunc()
 
 	// Revoke at the provider before deleting local credentials. If remote
 	// revocation fails, preserve both the token and DCR metadata for retry.
@@ -116,17 +123,27 @@ func revokeCommunityMode(ctx context.Context, app string) error {
 	if err != nil {
 		return fmt.Errorf("failed to load DCR client for provider revocation: %w", err)
 	}
-	token, err := getCommunityTokenFunc(ctx, app)
-	if err != nil {
-		return fmt.Errorf("failed to load OAuth token for provider revocation: %w", err)
-	}
-	if err := revokeProviderTokenFunc(ctx, dcrClient, token); err != nil {
-		return fmt.Errorf("failed to revoke OAuth access: %w", err)
+	providerRevoked := false
+	if dcrClient.RevocationEndpoint != "" {
+		token, err := getCommunityTokenFunc(ctx, app)
+		if err != nil {
+			return fmt.Errorf("failed to load OAuth token for provider revocation: %w", err)
+		}
+		if err := revokeProviderTokenFunc(ctx, dcrClient, token); err != nil {
+			return fmt.Errorf("failed to revoke OAuth access: %w", err)
+		}
+		providerRevoked = true
+	} else {
+		fmt.Printf("Note: OAuth provider for %s does not advertise revocation; deleting local token only\n", app)
 	}
 
-	// Delete OAuth token from docker pass only after provider revocation.
+	// Delete the local token after provider revocation succeeds, or immediately
+	// when the provider does not support remote revocation.
 	if err := deleteOAuthTokenFunc(ctx, appID); err != nil {
-		return fmt.Errorf("provider token revoked but failed to delete local OAuth token: %w", err)
+		if providerRevoked {
+			return fmt.Errorf("provider token revoked but failed to delete local OAuth token: %w", err)
+		}
+		return fmt.Errorf("failed to delete local OAuth token: %w", err)
 	}
 
 	// Delete DCR client from docker pass (soft failure -- entry may not exist
--- a/pkg/gateway/handlers.go
+++ b/pkg/gateway/handlers.go
@@ -54,13 +54,17 @@ func (g *Gateway) mcpToolHandler(serverName string, tool catalog.Tool) mcp.ToolH
 			attribute.String("mcp.server.image", tool.Container.Image),
 		}
 		ctx, span := telemetry.StartToolCallSpan(ctx, req.Params.Name, spanAttrs...)
-		defer span.End()
-		telemetry.ToolCallCounter.Add(ctx, 1, metric.WithAttributes(
+		metricAttrs := []attribute.KeyValue{
 			attribute.String("mcp.server.name", serverName),
 			attribute.String("mcp.server.type", "docker"),
 			attribute.String("mcp.tool.name", req.Params.Name),
 			attribute.String("mcp.client.name", clientName),
-		))
+		}
+		defer func() {
+			telemetry.ToolCallDuration.Record(ctx, float64(time.Since(startTime).Milliseconds()), metric.WithAttributes(metricAttrs...))
+			span.End()
+		}()
+		telemetry.ToolCallCounter.Add(ctx, 1, metric.WithAttributes(metricAttrs...))
 
 		if g.policyClient != nil {
 			policyReq := g.configuration.policyRequest(serverName, tool.Name, policy.ActionInvoke)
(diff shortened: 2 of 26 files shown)


pkg/oauth/revoke.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+		req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+		req.Header.Set("Accept", "application/json")
+
+		resp, err := httpClient.Do(req)
+		if err != nil {
+			return fmt.Errorf("revoking %s for %s: %w", item.hint, client.ServerName, err)
+		}
+		_ = resp.Body.Close()
+		if resp.StatusCode != http.StatusOK {
```
kgprs: RFC 7009 §2.2 says the server SHOULD respond with HTTP 200 but many implementations return `204 No Content` (empty body). This check will treat a valid 204 as a failure. Consider allowing the full 2xx range:
```go
if resp.StatusCode < 200 || resp.StatusCode >= 300 {
    return fmt.Errorf(...)
}
```

kgprs (commented):
## Review: Harden OAuth lifecycle and tool policy enforcement

Good security hardening work overall. The changes address a real set of vulnerabilities — key-collision in credential storage, cross-origin header leakage on redirects, missing issuer validation in OAuth discovery, and POCI tools bypassing policy/audit/telemetry. The test coverage for new behaviours is thorough.

A few points worth discussing before merge, noted as inline comments. Summary below.

### What's solid
- **Token key collision fix** (base64url-encoded versioned key + safe legacy migration) is well-designed and the test suite covers the attacker-controlled provider-name scenario cleanly.
- **Same-origin enforcement** for remote MCP headers is defence-in-depth at both the RoundTripper and CheckRedirect layers. The canonical host comparison (normalising default ports) is correct.
- **Issuer strict-equality check** is the right call per RFC 8414 — trailing slash differences correctly fail in the test.
- **POCI policy enforcement** brings container-image tools into parity with MCP server tools.
- **Audit + telemetry** wiring in `mcpToolHandler` is consistent with the existing server tool handler pattern.

### Concerns (see inline comments)
1. `RevokeTokenAtProvider` accepts only HTTP 200; RFC 7009 §2.2 permits 204 — some providers return empty body with 204.
2. No provider revocation endpoint = hard error, blocking local credential cleanup for decommissioned providers.
3. `revokeCEMode` now hard-fails but lacks a test for that failure path (community mode has one, CE mode does not).
4. Confidential client credentials are not sent with the revocation request despite `RevocationEndpointAuthMethodsSupported` being stored.
5. Policy denials in `mcpToolHandler` are not captured in the telemetry span (span is created after the policy block).

pkg/oauth/revoke.go
```diff
@@ -0,0 +1,81 @@
...
+	"golang.org/x/oauth2"
+
+	"github.com/docker/mcp-gateway/pkg/oauth/dcr"
+)
+
+// RevokeTokenAtProvider invalidates the refresh and access tokens at the
+// provider before callers remove their local copy. Redirects are deliberately
+// not followed because the request body contains bearer credentials.
+func RevokeTokenAtProvider(ctx context.Context, client dcr.Client, token *oauth2.Token) error {
+	if client.RevocationEndpoint == "" {
```
kgprs, line 28: When the OAuth provider doesn't advertise a revocation endpoint (common with older or simpler providers), this returns an error and leaves local credentials intact. That's the safe default, but it means users can't clean up credentials for a decommissioned or endpoint-less provider at all.

Consider adding a `--local-only` / `--force` flag on the CLI `docker mcp oauth revoke` command that skips provider-side revocation and just removes local credentials. The current message even hints at this (`"local token was preserved"`) but provides no path for the user to act on it.

cmd/docker-mcp/oauth/revoke.go
```diff
@@ -84,10 +87,10 @@ func revokeCEMode(ctx context.Context, app string) error {
 	credHelper := pkgoauth.NewReadWriteCredentialHelper()
 	manager := pkgoauth.NewManager(credHelper)
 
-	// Delete OAuth token
+	// Revoke at the provider before deleting local credentials. If remote
+	// revocation fails, preserve both the token and DCR metadata for retry.
 	if err := manager.RevokeToken(ctx, app); err != nil {
-		// Token might not exist, continue to DCR deletion
-		fmt.Printf("Note: %v\n", err)
+		return fmt.Errorf("failed to revoke OAuth access: %w", err)
```
kgprs, line 101: This hard-fails CE mode revocation if `manager.RevokeToken` returns any error — provider unreachable, token already expired at the provider, token never stored, etc. The community-mode path has `TestRevokeCommunityMode_PreservesLocalCredentialsWhenProviderRevokeFails` covering the equivalent behaviour, but there's no matching test for CE mode here.

A test along the lines of:
```go
func TestRevokeCEMode_FailsHardWhenProviderRevocationFails(t *testing.T) { ... }
```
would cover the regression where this reverts to soft-fail.

pkg/oauth/revoke.go
```diff
@@ -0,0 +1,81 @@
...
+		return http.ErrUseLastResponse
+	}
+
+	for _, item := range requests {
+		form := url.Values{
+			"token":           {item.token},
+			"token_type_hint": {item.hint},
+		}
+		if client.ClientID != "" {
+			form.Set("client_id", client.ClientID)
```
kgprs, line 69: Only `client_id` is sent; no client secret or other credentials. RFC 7009 §2.1 requires confidential clients to authenticate to the revocation endpoint. The DCR client struct now carries `RevocationEndpointAuthMethodsSupported` but it's not used here.

For purely public clients (no secret) this is fine. If any registered clients are confidential, providers requiring authentication will return `401`. Worth documenting the public-client-only assumption in the function comment, or wiring `client.ClientSecret` (if stored) into the form as `client_secret`.

pkg/gateway/handlers.go
```diff
@@ -41,12 +41,47 @@ func inferServerTransportType(serverConfig *catalog.ServerConfig) string {
...
+			policyReq := g.configuration.policyRequest(serverName, tool.Name, policy.ActionInvoke)
+			decision, err := g.policyClient.Evaluate(ctx, policyReq)
+			event := buildAuditEvent(policyReq, decision, err, auditClientInfoFromSession(req.Session))
+			submitAuditEvent(g.policyClient, event)
+			if err != nil {
+				telemetry.RecordToolError(ctx, nil, serverName, "docker", req.Params.Name)
+				return nil, fmt.Errorf("policy check failed for %s/%s: %w", serverName, tool.Name, err)
+			}
+			if !decision.Allowed {
+				return nil, fmt.Errorf("policy denied tool %s on server %s: %s", tool.Name, serverName, decision.Reason)
```
kgprs: The policy check runs before `StartToolCallSpan` (~line 63), so a policy denial exits without ever creating a span. Policy-blocked calls will be absent from distributed traces entirely — only the audit log captures them.

If you want policy denials visible in traces, either start the span before this block and set its status, or at minimum call `telemetry.RecordToolError` with a non-nil span by hoisting `StartToolCallSpan` above the policy check.

pkg/oauth/token_store.go
```diff
@@ -12,6 +13,53 @@ import (
...
+	username, secret, err := helper.Get(legacyKey)
+	if err != nil {
+		return "", err
+	}
+
+	// Transparently migrate unambiguous legacy entries. Migration must finish
+	// before the token is returned so a slash-bearing provider cannot read the
+	// same legacy slot through an older ambiguous key path in this process.
+	if err := helper.Add(&credentials.Credentials{ServerURL: key, Username: username, Secret: secret}); err != nil {
+		return "", fmt.Errorf("migrating OAuth token credential: %w", err)
```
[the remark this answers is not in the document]
kgprs: Fixed in `b228b2e5`. Read-only `CredentialHelper` call paths now read an unambiguous legacy secret without attempting `Add`/`Delete`; writable `TokenStore` retrieval still performs the migration. `TestGetTokenCredentialReadsLegacyKeyWithoutMigratingThroughReadOnlyHelper` covers the upgrade path.

pkg/oauth/revoke.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/docker/mcp-gateway/pkg/oauth/dcr"
+)
+
+// RevokeTokenAtProvider invalidates the refresh and access tokens at the
+// provider before callers remove their local copy. Redirects are deliberately
+// not followed because the request body contains bearer credentials.
+func RevokeTokenAtProvider(ctx context.Context, client dcr.Client, token *oauth2.Token) error {
+	if client.RevocationEndpoint == "" {
+		return fmt.Errorf("OAuth provider for %s does not advertise a revocation endpoint; local token was preserved", client.ServerName)
```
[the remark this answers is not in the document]
kgprs: Fixed in `b228b2e5`. A missing revocation endpoint, including on legacy DCR records, now takes the explicit local-cleanup path: it logs a note and deletes the local token instead of trapping the credential. When an endpoint is present, remote failure still preserves local state. Manager and community-mode tests cover the behavior.

pkg/oauth/revoke.go
kgprs: Fixed in `b228b2e5`. Provider revocation now accepts the full HTTP 2xx range, and the provider test returns `204 No Content` to cover this case.

pkg/oauth/revoke.go
kgprs: Fixed in `b228b2e5`. Providers without revocation support now use a local-cleanup path automatically, with a user-visible note, while providers that advertise an endpoint still fail closed and preserve local credentials on remote errors. This avoids requiring a separate force flag for the endpoint-less case.

cmd/docker-mcp/oauth/revoke.go
kgprs: Fixed in `b228b2e5`. Added `TestRevokeCEMode_FailsHardWhenProviderRevocationFails` through a narrow manager seam; it verifies the provider error is returned and the DCR client is not deleted.

pkg/oauth/revoke.go
kgprs: Documented in `b228b2e5`. This implementation only creates public DCR clients: registration always requests `token_endpoint_auth_method=none`, and no client secret is stored. The function comment now states that revocation identifies the public client with `client_id` only and does not support confidential-client authentication.

pkg/gateway/handlers.go
kgprs: Fixed in `b228b2e5`. The POCI tool-call span and counter now start before policy evaluation; evaluation failures and denials record an error against the non-nil span and set error status. `TestPOCIPolicyDenialCreatesErrorSpan` verifies the denied-call trace.

cmd/docker-mcp/oauth/revoke.go
```diff
@@ -109,10 +119,31 @@ func revokeCommunityMode(ctx context.Context, app string) error {
 		return fmt.Errorf("invalid server name %q: %w", app, err)
 	}
 
-	// Delete OAuth token from docker pass
+	dcrClient, err := getCommunityDCRClientFunc(ctx, app)
+	if err != nil {
+		return fmt.Errorf("failed to load DCR client for provider revocation: %w", err)
```
arielho-docker, line 137: Before this PR, `revokeCommunityMode` went straight to `deleteOAuthTokenFunc` and treated a missing token/DCR entry as a soft no-op (`Note: %v`). Now a failed DCR-client lookup hard-fails the whole revoke call.

That means `docker mcp oauth revoke <server>` on a server that was never OAuth-authorized (no DCR registration exists) now returns an error instead of completing as a no-op. This breaks idempotent/defensive use of `revoke` (e.g. a cleanup script that revokes unconditionally across a catalog).

None of the new tests cover this "no DCR client" path — all three new `TestRevokeCommunityMode_*` tests mock `getCommunityDCRClientFunc` to succeed. Was this UX change intentional? If not, consider short-circuiting to a graceful "nothing to revoke" when this lookup fails with a not-found error.

pkg/gateway/handlers.go
```diff
@@ -41,12 +41,50 @@ func inferServerTransportType(serverConfig *catalog.ServerConfig) string {
...
+			submitAuditEvent(g.policyClient, event)
+			if err != nil {
+				telemetry.RecordToolError(ctx, span, serverName, "docker", req.Params.Name)
+				span.SetStatus(codes.Error, "Policy evaluation failed")
+				return nil, fmt.Errorf("policy check failed for %s/%s: %w", serverName, tool.Name, err)
+			}
+			if !decision.Allowed {
+				telemetry.RecordToolError(ctx, span, serverName, "docker", req.Params.Name)
+				span.SetStatus(codes.Error, "Policy denied tool call")
+				return nil, fmt.Errorf("policy denied tool %s on server %s: %s", tool.Name, serverName, decision.Reason)
```
arielho-docker, line 82: `telemetry.ToolCallDuration.Record(...)` only runs after `runToolContainer` returns (line ~97 below), so this early return (policy denial) — and the unmarshal-error return above — never record a duration sample, only the counter and error span/metric.

The PR description says the goal was to record "the existing tool-call counter, duration, error metric, and trace status" for the POCI path, so duration coverage is currently incomplete for failure paths. Low severity, but worth a follow-up if duration-based alerting needs to account for error-path latency too.

arielho-docker (commented):
Reviewed alongside the enterprise companion PR (docker/mcp-gateway-enterprise#2548). Overall this is solid, well-tested security hardening. Two inline notes below; nothing blocking.

pkg/gateway/handlers.go
kgprs: Addressed in aa94f45f: the existing tool-duration histogram is now recorded by a handler-entry defer, so policy evaluation errors, policy denials, argument decode failures, and container results each emit exactly one duration sample. Tests assert the duration metric for both a POCI policy denial and malformed arguments, and the metric documentation now reflects end-to-end handling latency. Validation: `go test -count=1 ./pkg/gateway ./pkg/telemetry`, `make lint`, and `make test`.

cmd/docker-mcp/oauth/revoke.go
kgprs: Addressed in aa94f45f: DCR lookup now preserves the not-found sentinel, and community revoke treats only that case as idempotent. It skips provider calls, performs best-effort local token/DCR and stale Desktop cleanup, and returns success; other lookup failures still fail loudly. The regression test covers a never-authorized server with missing local entries. Validation: `go test -count=1 ./cmd/docker-mcp/oauth ./pkg/oauth`, `make lint`, and `make test`.

arielho-docker (approved):

PR #547, opened 2026-08-11
fix: preserve Windows drive roots in Docker binds

### LINKED ISSUE #541: Windows: bare drive-root server path (D:\) is cleaned to "d:", producing a volume spec the daemon rejects; gateway reports only 'initialize: EOF'
### Summary

On Windows, a server whose configured path is a bare drive root (e.g. `D:\`) produces a volume spec the Docker daemon rejects, so the server container never starts. The gateway surfaces only a generic `initialize: EOF`, with no hint that the mount was the cause.

The root cause is in `cleanDockerHostPath` ([pkg/gateway/docker_binds.go](https://github.com/docker/mcp-gateway/blob/main/pkg/gateway/docker_binds.go)):

```go
func cleanDockerHostPath(p string) (string, error) {
	p = strings.TrimSpace(strings.ReplaceAll(p, "\\", "/"))
	if isWindowsAbsPath(p) {
		return strings.ToLower(path.Clean(p)), nil
	}
	...
```

`path.Clean("d:/")` strips the trailing slash and returns `"d:"`. For any deeper path (`D:\foo` → `d:/foo`) this is harmless, but for a drive root the trailing separator is the only thing marking it as a path — the cleaned `d:` then gets reassembled into a bind spec like `d::/D/:rw`, which the daemon rejects with `The system cannot find the file specified.`

### Environment

- Windows 11 Pro, Docker Desktop, engine 29.6.2
- `docker mcp version`: v0.43.1
- Verified 2026-07-28

### Reproduction

1. The gateway's cleaned form fails at the daemon (container never runs):

```
> docker run --rm --entrypoint ls -v "d::/D/:rw" mcp/filesystem /D/
docker: Error response from daemon: The system cannot find the file specified.
```

2. The same drive root mounts fine when the separator is preserved — both the native Windows spec and the slash form work:

```
> docker run --rm --entrypoint ls -v 'D:\:/D:ro' mcp/filesystem /D
$RECYCLE.BIN
Caches
...

> docker run --rm --entrypoint ls -v "d:/:/D:rw" mcp/filesystem /D
$RECYCLE.BIN
Caches
...
```

This shows drive roots are mountable and the gateway's cleaned form is specifically at fault.

Through the gateway, the observable symptom is just:

```
starting client: calling "initialize": EOF
```

### Suggested fix

In `cleanDockerHostPath`, preserve the trailing separator when the cleaned Windows-absolute path is a bare drive root — i.e. emit `d:/` (or `D:\`) instead of `d:`. Something like:

```go
if isWindowsAbsPath(p) {
	cleaned := strings.ToLower(path.Clean(p))
	if len(cleaned) == 2 && cleaned[1] == ':' {
		cleaned += "/"
	}
	return cleaned, nil
}
```

### Secondary issue: diagnostics

A bad volume on one server aborts with `initialize: EOF` and nothing pointing at the mount. Surfacing the daemon's error (or at least the failing `-v` spec) for a container that dies before initialize would make this class of misconfiguration self-diagnosing. Possibly related territory: #537 also concerns `cleanDockerHostPath` normalization on Windows changing the meaning of bind sources.

shixi-li:
**What I did**

- Preserve the trailing slash when a Windows absolute host path normalizes to a bare drive root.
- Add cross-platform regression coverage for both slash forms, non-root paths, and the final Docker bind string.

**Why**

`path.Clean("D:/")` returns `"D:"`. The gateway then rebuilt the volume as `d::/D:ro`, which Docker rejects. Keeping the root separator produces the valid `d:/:/D:ro` form without changing non-root Windows paths or UNC handling.

**Related issue**

Fixes #541

**Validation**

- `go test ./pkg/gateway -run 'TestCleanDockerHostPathPreservesWindowsDriveRoot|TestNormalizeDockerVolumeBindPreservesWindowsDriveRoot' -count=1`
- `go test ./... -short -count=1`
- `go vet ./...`

### DIFF
--- a/pkg/gateway/docker_binds.go
+++ b/pkg/gateway/docker_binds.go
@@ -201,7 +201,11 @@ func isWindowsAbsPath(p string) bool {
 func cleanDockerHostPath(p string) (string, error) {
 	p = strings.TrimSpace(strings.ReplaceAll(p, "\\", "/"))
 	if isWindowsAbsPath(p) {
-		return strings.ToLower(path.Clean(p)), nil
+		cleaned := strings.ToLower(path.Clean(p))
+		if len(cleaned) == 2 && cleaned[1] == ':' {
+			cleaned += "/"
+		}
+		return cleaned, nil
 	}
 	if strings.HasPrefix(p, "//") {
 		return "//" + path.Clean(strings.TrimPrefix(p, "//")), nil
--- a/pkg/gateway/docker_binds_test.go
+++ b/pkg/gateway/docker_binds_test.go
@@ -1,0 +1,34 @@
+package gateway
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestCleanDockerHostPathPreservesWindowsDriveRoot(t *testing.T) {
+	for _, tc := range []struct {
+		input string
+		want  string
+	}{
+		{input: `D:\`, want: "d:/"},
+		{input: "D:/", want: "d:/"},
+		{input: `D:\data`, want: "d:/data"},
+	} {
+		t.Run(tc.input, func(t *testing.T) {
+			got, err := cleanDockerHostPath(tc.input)
+			require.NoError(t, err)
+			require.Equal(t, tc.want, got)
+		})
+	}
+}
+
+func TestNormalizeDockerVolumeBindPreservesWindowsDriveRoot(t *testing.T) {
+	got, err := normalizeDockerVolumeBindWithRoots(
+		`D:\:/D:ro`,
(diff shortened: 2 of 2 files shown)


shixi-li:
Hi maintainers — the fork workflows for this PR are currently awaiting approval: [CI](https://github.com/docker/mcp-gateway/actions/runs/31456853224), [validate](https://github.com/docker/mcp-gateway/actions/runs/31456853245), and [CodeQL](https://github.com/docker/mcp-gateway/actions/runs/31456853272). The local validation results are included in the PR description. When convenient, could you approve these workflows so the change can be evaluated? Thanks!

Issue #10, opened 2025-06-18
sseEndpoint support in catalog


slimslenderslacks:
To align with the v2 registry concept of a "remote", we could adopt their Remote type here:

```
   Remote:
      type: object
      required:
        - transport_type
        - url
      properties:
        transport_type:
          type: string
          enum: [streamable, sse]
          example: "sse"
        url:
          type: string
          format: uri
          example: "https://mcp-fs.example.com/sse"
        headers:
          type: array
          items:
            $ref: '#/components/schemas/KeyValueInput'
```

We could have an array of `remotes: []` too.  Using the name `sseEndpoint` makes it seem like we'd have to have one property for every future transport_type, and `sse` is now deprecated.

PR #549, opened 2026-08-11
Harden OAuth discovery and tool dispatch

kgprs:
## Context

This is a follow-up to #546. It closes the security, authorization, and observability gaps found during the post-merge security review: credential-bearing OAuth requests could still follow redirects; protected-resource discovery was not fully bound to the MCP resource; `ActionInvoke` enforcement was duplicated across dispatch paths; CE revocation could not remove stale DCR state after its local token disappeared; Windows browser launch crossed a `cmd.exe` shell boundary with an untrusted OAuth URL; case-insensitive tool allowlisting could enable a case-variant tool the operator did not select; code mode rebuilt an unfiltered backend tool surface outside the gateway's registration boundary; globally registered code-mode tools captured the creator's session identity; and self-hosted dynamic management treated a noop policy as authorization to activate catalog-only servers.

### Refuse redirects for credential-bearing OAuth requests

- Build the guarded OAuth client through the standard direct or Docker Desktop proxy constructors.
- Export and reuse one credential-bearing OAuth client across CE and community authorization-code exchange, CE and community refresh, and provider revocation.
- Refuse every redirect in that shared client so authorization codes, PKCE verifiers, refresh tokens, and credential headers are never replayed to a redirect target.
- Keep redirect responses visible to the OAuth caller so the operation fails instead of silently following them.

### Open OAuth authorization URLs without a shell

- Validate authorization URLs before launching a platform browser handler: require an absolute public HTTPS URL by default and reject userinfo or unsafe hosts.
- Replace Windows `cmd /c start` with `rundll32.exe url.dll,FileProtocolHandler`, passing the complete authorization URL as one process argument.
- Preserve the explicit local-development opt-in for HTTP/private authorization URLs through `DOCKER_MCP_ALLOW_INSECURE_REMOTE_URLS=1`.

### Bind OAuth discovery to the requested MCP resource

- Require a `resource_metadata` URL from `WWW-Authenticate` to have the same scheme, host, and effective port as the MCP server.
- Reapply that same-origin requirement to every metadata redirect hop.
- Require protected-resource metadata's `resource` value to exactly match the full MCP server URL being probed, including path and query.
- Construct the fallback protected-resource well-known URL using RFC 9728's path-aware form.
- Consume `mcp-gateway-oauth-helpers v0.0.7` and regenerate the vendored module from that release.
- Apply the same resource, origin, redirect, and issuer checks to the vendored helper implementation.
- Guard helper authorization-server metadata fetches at dial time: require public HTTPS destinations, reject private and reserved DNS results, pin validated IPs, and enforce the guard across redirects.
- Continue requiring exact RFC 8414 issuer equality. A trailing-slash difference is intentionally a mismatch.

### Enforce exact-case tool allowlisting

- Match bare, server-scoped, and image-scoped tool selectors using the exact case advertised by the MCP server, consistent with MCP tool dispatch.
- Keep global, server, and image wildcards exact (`*`, `server:*`, and `image:*`).
- Reject case-folded collisions between registered tool names, reserved gateway tools, and existing tool registrations before they can create ambiguous operator intent.
- Warn on the same case-folded collisions while assembling catalog tools so configuration problems are visible before invocation.
- Keep collision detection scoped to tools: prompts and resources remain independent MCP capability namespaces, and resource URIs retain their original case.

### Keep code mode inside the filtered, session-scoped registration boundary

- Stop re-listing backend tools and creating separate direct-session handlers when a code-mode tool is created or executed.
- Build each selected server's JavaScript functions from `g.toolRegistrations`, which contains the exact-case allowlisted and `ActionLoad`-filtered tool definitions.
- Reuse each registered handler so code mode inherits the same `ActionInvoke` middleware, telemetry, original backend tool-name mapping, and prefixed gateway tool name as direct dispatch and `mcp-exec`.
- Remove the creator session and request metadata from the globally registered code-mode adapter.
- Propagate `Session`, `Extra`, and request `Meta` dynamically from every outer code-mode invocation to its nested registered-handler calls, so the backend connection, roots/state, policy identity, and audit attribution always belong to the current caller.
- Validate requested servers against the exact enabled-server list rather than the full catalog, preventing profile catalog entries and case variants from being selected before activation.
- Snapshot registrations under the capability lock and sort them by exact tool name for deterministic generated documentation and function installation.

### Require authoritative policy for catalog-only management

- Treat the configured registry/server list as the operator-enabled management boundary.
- Let `mcp-add`, `mcp-config-set`, and profile activation skip new authorization only for servers already inside the operator-enabled set.
- Require a real policy provider and an allowed `ActionLoad` decision before an individual tool or profile can activate or configure a catalog-only server.
- Build profile activation decisions from the profile's server metadata before merging it into the gateway, preserving complete policy inputs without weakening the active-set boundary.
- Treat a nil policy client and the self-hosted `policy.NoopClient` as absence of authorization, rather than an allow decision.
- Reuse the shared load-policy gate so authoritative allow/deny/error decisions retain the existing audit event and fail-closed behavior.
- Return an error before validation, image pulls, capability reloads, or configuration mutation when profile access is not authorized.

### Centralize `ActionInvoke` enforcement

- Add one fail-closed `ActionInvoke` middleware and attach it when backend tool handlers are registered.
- Route direct MCP calls, POCI tools, `mcp-exec`, and code mode through the same wrapped handlers.
- Preserve each path's existing client-facing error shape while producing one policy decision and one audit event per attempted invocation.
- Keep both POCI and MCP-server telemetry outside the policy middleware so denials and evaluator failures record the normal tool-call counter, end-to-end duration, error counter, and error span.
- Remove the duplicate MCP-server instrumentation from the backend handler, ensuring allowed calls and policy short-circuits each produce exactly one telemetry envelope.

### Make missing-token CE revocation idempotent

- Preserve the credential-helper not-found sentinel through `TokenStore.Retrieve`.
- Treat only that sentinel as an already-completed local token deletion and continue removing the DCR client.
- Continue failing closed for provider, transport, malformed-token, and other credential-helper failures.

## Compatibility notes

- Protected-resource and issuer comparisons use exact string equality. Providers whose metadata changes only by a trailing slash will be rejected and must publish the exact identifier used for discovery.
- Explicit `resource_metadata` failures are no longer ignored. A server advertising invalid, cross-origin, unreachable, or mismatched metadata fails discovery before browser authorization or code exchange.
- Helper authorization-server metadata discovery permits only public HTTPS destinations by default. Local, private-network, or HTTP OAuth requires the explicit development opt-in `DOCKER_MCP_ALLOW_INSECURE_REMOTE_URLS=1`; this disables the network guard and must not be used with untrusted MCP servers.
- `--open-browser` applies the same URL policy before starting the platform handler. Without the development opt-in, relative, non-HTTP(S), HTTP, userinfo-bearing, localhost, private, and reserved authorization URLs are rejected.
- Tool allowlists are now case-sensitive. Existing selectors whose capitalization differs from the advertised MCP tool name must be corrected; wildcards retain their existing behavior.
- Code mode now exposes only tools registered for an enabled server. Disabled, `ActionLoad`-denied, catalog-only, and case-mismatched tools no longer appear as JavaScript functions; prefixed tools must be called by their registered gateway name.
- Globally registered code-mode tools now use the invoking client's session and request identity on every call; they no longer retain the session that created the generated tool.
- In self-hosted/no-governance mode, dynamic tools and profile activation cannot add or configure a catalog-only server. Pre-enable the server in the registry, use a fixed `--servers` set, or configure an enforcing governance policy that allows its `ActionLoad` request.
- CE revoke remains strict for real provider and storage failures; only an already-absent local token permits DCR cleanup to continue.

## Automated validation

- `make format`
- `go test ./pkg/oauth ./pkg/oauthdiscovery ./cmd/docker-mcp/oauth ./pkg/gateway`
- `go test -race -count=1 ./pkg/gateway -run 'TestCodeModeUsesRegisteredServerTools|TestCodeModeRejectsServersOutsideEnabledSet|TestCheckServerManagementAccess|TestNoopPolicyCannotMutateCatalogOnlyServer|TestActivateProfileCannotActivateCatalogOnlyServerWithoutEnforcingPolicy|TestMCPServerPolicyFailuresRecordCompleteTelemetry'`
- `go vet ./pkg/codemode ./pkg/gateway`
- `go test github.com/docker/mcp-gateway-oauth-helpers`
- `make test`
- `make lint`

## Manual testing

Use a test MCP/OAuth provider and a credential-capture HTTP server. Point a catalog entry at the test MCP server, then run the branch build through the normal CE/community test setup.

1. **Token endpoint redirect does not leak credentials**
   - Configure token endpoint A to return `307 Temporary Redirect` with `Location` set to capture endpoint B on another host or port.
   - Complete authorization through the point where the gateway exchanges the authorization code, once in CE mode and once through the Desktop community-server path.
   - Repeat with both `307 Temporary Redirect` and `308 Permanent Redirect`; confirm each exchange fails on the redirect response.
   - Confirm endpoint B received no request, `Authorization` header, client ID, client secret, authorization code, PKCE verifier, or refresh token.

2. **Sibling-origin `resource_metadata` is rejected before fetch**
   - Have the MCP endpoint return `401` with `WWW-Authenticate: Bearer resource_metadata="https://metadata.example.test/.well-known/oauth-protected-resource"` while the probed server is `https://mcp.example.test/mcp`.
   - Run OAuth discovery.
   - Confirm discovery reports a same-origin validation error and the metadata host receives no request.

3. **Cross-origin metadata redirects are rejected**
   - Advertise a same-origin metadata URL from the MCP endpoint.
   - Make that URL return a redirect to a second host or port.
   - Run OAuth discovery and confirm it fails at the redirect.
   - Confirm the redirect target receives no request.

4. **Protected-resource identity is exact**
   - Probe `https://mcp.example.test/mcp` and return same-origin protected-resource metadata with `"resource": "https://mcp.example.test/other"`.
   - Confirm discovery fails before DCR, browser authorization, or token exchange.
   - Change `resource` to exactly `https://mcp.example.test/mcp`; confirm discovery continues.
   - Repeat with only a trailing slash or query difference and confirm it is rejected.

5. **Authorization-server issuer is exact in both implementations**
   - Request metadata for issuer `https://auth.example.test` while returning `"issuer": "https://auth.example.test/"`, then repeat with an unrelated issuer.
   - Confirm both cases fail before the authorization URL is used.
   - Return the exact requested issuer and confirm discovery succeeds.
   - Exercise both the gateway discovery package and the vendored helper call path.

6. **Every tool dispatch path uses the shared policy gate**
   - Configure the policy evaluator to deny one `(server, tool, invoke)` tuple and enable backend invocation logging.
   - Attempt that tool directly, through a POCI registration, through `mcp-exec`, and through code mode.
   - Confirm all four attempts are denied and the backend/container is never invoked.
   - Confirm each attempt creates exactly one policy evaluation and one audit event.
   - For both POCI and MCP-server attempts, confirm the tool call counter, duration histogram, error counter, and error span are each recorded exactly once even though the backend handler does not run.
   - Repeat with an evaluator error and confirm all paths fail closed with the same telemetry envelope and exactly one audit event.

7. **CE revoke cleans stale DCR state when the token is already absent**
   - Authorize a CE OAuth server so both its token credential and DCR client exist.
   - Delete only the token entry (`https://oauth-token.mcp/v2/...`) from the configured credential helper, leaving the DCR entry intact.
   - Run `docker mcp oauth revoke <server>`.
   - Confirm the command reports that the token is already absent, exits successfully, and removes the DCR client.
   - As a negative control, reauthorize, make the provider revocation endpoint return HTTP 500, and run revoke again. Confirm the command exits non-zero and preserves the local token and DCR client for retry.

8. **Local OAuth requires explicit opt-in**
   - Configure a localhost MCP server that discovers an HTTP localhost OAuth issuer.
   - Unset `DOCKER_MCP_ALLOW_INSECURE_REMOTE_URLS` and run OAuth discovery. Confirm it fails before dialing authorization-server metadata.
   - Set `DOCKER_MCP_ALLOW_INSECURE_REMOTE_URLS=1` and repeat. Confirm discovery reaches the local provider and the OAuth flow can proceed.
   - Unset the environment variable after testing.

9. **Windows browser launch treats the authorization URL as data**
   - On Windows, configure an authorization endpoint whose query includes `x=1&calc.exe&y=` and run OAuth authorization with `--open-browser`.
   - Confirm the default browser receives the complete URL and no `calc.exe` process or other extra command is launched.
   - Repeat with a relative URL, `file:` URL, and plain HTTP public URL while `DOCKER_MCP_ALLOW_INSECURE_REMOTE_URLS` is unset. Confirm each is rejected before a browser handler starts.
   - As a local-development control, set `DOCKER_MCP_ALLOW_INSECURE_REMOTE_URLS=1` and confirm an HTTP localhost authorization URL can still be opened.

10. **Tool allowlists honor exact MCP tool names**
   - Configure a test server to advertise two distinct tools, `ReadFile` and `readfile`.
   - Allow only `ReadFile` with a bare selector. Confirm `ReadFile` is exposed and callable while `readfile` is hidden and cannot be dispatched through the gateway.
   - Repeat with an exact server-scoped selector and an exact image-scoped selector; confirm a case-only mismatch does not enable either tool.
   - Confirm exact `*`, `server:*`, and `image:*` wildcards still expose the intended tools.
   - Attempt to register a tool whose name differs only by case from a reserved gateway tool or an existing registered tool. Confirm registration is rejected and catalog assembly reports a collision warning.
   - As namespace controls, expose a tool and prompt with the same name and confirm both remain usable through their separate MCP methods; expose two resource URIs differing only by path case and confirm they remain distinct.

11. **Code mode cannot restore filtered tools or catalog-only servers**
   - Configure an enabled test server to advertise `ReadFile`, `readfile`, and `AdminDelete`. Allowlist only `ReadFile` and configure the policy evaluator to deny `AdminDelete` at `ActionLoad`.
   - Create a code-mode tool for that server. Confirm its generated JavaScript documentation contains only `ReadFile` and does not contain `readfile` or `AdminDelete`.
   - Execute a script that calls `ReadFile` and confirm the already-registered handler runs with normal telemetry and exactly one `ActionInvoke` evaluation.
   - Attempt scripts that call `readfile` and `AdminDelete`. Confirm both functions are undefined and neither reaches the backend or produces an invocation-policy audit event.
   - In profile mode, leave a second server in the catalog but do not activate it. Attempt to create code mode for that catalog-only name and for a case variant of the enabled server name; confirm both requests fail as not enabled before any backend client is acquired.
   - Activate the second server and repeat with its exact name. Confirm code mode then exposes only that server's registered, filtered tool set.

12. **Generated code-mode tools preserve each invoking session's identity**
   - Open MCP session A and create a code-mode tool for an enabled backend server; keep the generated tool registered.
   - Invoke that generated tool from session A and record the backend client/session, roots or session-scoped state, request metadata, and audit identity.
   - Open session B and invoke the same globally registered generated tool, both directly and through `mcp-exec` if that path is exposed by the client.
   - Confirm both session-B calls use B's backend connection, roots/state, request metadata, policy identity, and audit attribution; none of A's identity or state is visible.
   - Invoke the tool from A again and confirm it still uses A, demonstrating that identity is derived per invocation rather than captured when the tool was created.

13. **Noop policy cannot activate or configure catalog-only servers**
   - In a self-hosted/no-governance gateway, put server A in the registry and leave server B only in the catalog.
   - Run `mcp-add` for B and `mcp-config-set` for B. Confirm both return an MCP error, B is not appended to the enabled-server list, no B configuration is written, and no reload occurs.
   - Create a profile containing B and activate it with the nil/noop policy client. Confirm activation fails before image pulls or capability reloads and leaves B absent from the enabled-server list and configuration.
   - Run the same management operation for already-enabled A and confirm it retains the supported self-hosted behavior.
   - Configure an enforcing policy provider, allow B's `ActionLoad` request, and repeat both individual and profile activation. Confirm each succeeds and emits the existing policy audit event using B's complete server metadata.
   - Change the policy to deny B (and separately make evaluation fail), then repeat. Confirm each attempt fails closed and leaves the enabled-server list and configuration unchanged.

### DIFF
--- a/pkg/gateway/capabilitites.go
+++ b/pkg/gateway/capabilitites.go
@@ -177,10 +177,13 @@ func (g *Gateway) listCapabilities(ctx context.Context, serverNames []string, cl
 						capabilities.Tools = append(capabilities.Tools, ToolRegistration{
 							ServerName: serverConfig.Name,
 							Tool:       &prefixedTool,
-							Handler: g.withInvokePolicy(
-								serverConfig.Name,
-								tool.Name,
-								g.mcpServerToolHandler(serverConfig.Name, g.mcpServer, tool.Annotations, tool.Name),
+							Handler: withMCPServerPolicyErrorTelemetry(
+								serverConfig,
+								g.withInvokePolicy(
+									serverConfig.Name,
+									tool.Name,
+									g.mcpServerToolHandler(serverConfig.Name, g.mcpServer, tool.Annotations, tool.Name),
+								),
 							),
 						})
 					}
--- a/pkg/gateway/capabilitites.go
+++ b/pkg/gateway/capabilitites.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"runtime"
 	"slices"
-	"strings"
 	"sync"
 
 	"github.com/google/jsonschema-go/jsonschema"
@@ -358,18 +357,17 @@ func isToolEnabled(configuration Configuration, serverName, serverImage, toolNam
 
 	for _, enabled := range enabledTools {
 		if enabled == "*" ||
-			strings.EqualFold(enabled, toolName) ||
-			strings.EqualFold(enabled, serverName+":"+toolName) ||
-			strings.EqualFold(enabled, serverName+":*") ||
-			strings.EqualFold(enabled, "*") {
+			enabled == toolName ||
+			enabled == serverName+":"+toolName ||
+			enabled == serverName+":*" {
 			return true
 		}
 	}
 
 	if serverImage != "" {
 		for _, enabled := range enabledTools {
-			if strings.EqualFold(enabled, serverImage+":"+toolName) ||
-				strings.EqualFold(enabled, serverImage+":*") {
+			if enabled == serverImage+":"+toolName ||
+				enabled == serverImage+":*" {
 				return true
 			}
 		}
--- a/pkg/gateway/capabilitites.go
+++ b/pkg/gateway/capabilitites.go
@@ -176,7 +176,7 @@ func (g *Gateway) listCapabilities(ctx context.Context, serverNames []string, cl
 						capabilities.Tools = append(capabilities.Tools, ToolRegistration{
 							ServerName: serverConfig.Name,
 							Tool:       &prefixedTool,
-							Handler: withMCPServerPolicyErrorTelemetry(
+							Handler: withMCPServerToolTelemetry(
 								serverConfig,
 								g.withInvokePolicy(
 									serverConfig.Name,
--- a/pkg/gateway/capabilitites.go
+++ b/pkg/gateway/capabilitites.go
@@ -177,7 +177,11 @@ func (g *Gateway) listCapabilities(ctx context.Context, serverNames []string, cl
 						capabilities.Tools = append(capabilities.Tools, ToolRegistration{
 							ServerName: serverConfig.Name,
 							Tool:       &prefixedTool,
-							Handler:    g.mcpServerToolHandler(serverConfig.Name, g.mcpServer, tool.Annotations, tool.Name),
+							Handler: g.withInvokePolicy(
+								serverConfig.Name,
+								tool.Name,
+								g.mcpServerToolHandler(serverConfig.Name, g.mcpServer, tool.Annotations, tool.Name),
+							),
 						})
 					}
 				}
@@ -285,7 +289,11 @@ func (g *Gateway) listCapabilities(ctx context.Context, serverNames []string, cl
 				capabilities.Tools = append(capabilities.Tools, ToolRegistration{
 					ServerName: serverName,
 					Tool:       &mcpTool,
-					Handler:    g.mcpToolHandler(serverName, tool),
+					Handler: g.withPOCIToolTelemetry(
+						serverName,
+						tool,
+						g.withInvokePolicy(serverName, tool.Name, g.pociToolHandler(tool)),
+					),
 				})
 			}
 
--- a/pkg/gateway/tool_policy.go
+++ b/pkg/gateway/tool_policy.go
@@ -2,10 +2,13 @@ package gateway
 
 import (
 	"context"
-	"errors"
 	"fmt"
+	"time"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/metric"
 
 	"github.com/docker/mcp-gateway/pkg/catalog"
 	"github.com/docker/mcp-gateway/pkg/policy"
@@ -53,17 +56,52 @@ func (g *Gateway) withInvokePolicy(serverName, toolName string, next mcp.ToolHan
 	}
 }
 
-// withMCPServerPolicyErrorTelemetry preserves the error signal for policy
-// evaluator failures that occur before the MCP server handler starts its tool
-// telemetry. The server handler remains responsible for execution failures,
-// while POCI registrations use their existing outer telemetry middleware.
-func withMCPServerPolicyErrorTelemetry(serverConfig *catalog.ServerConfig, next mcp.ToolHandler) mcp.ToolHandler {
+// withMCPServerToolTelemetry records the complete tool-call telemetry envelope
+// outside policy enforcement, so policy denials and evaluator failures are
+// observable without double-counting calls that reach the backend handler.
+func withMCPServerToolTelemetry(serverConfig *catalog.ServerConfig, next mcp.ToolHandler) mcp.ToolHandler {
 	return func(ctx context.Context, req *mcp.CallToolRequest) (*mcp.CallToolResult, error) {
-		result, err := next(ctx, req)
-		var policyErr *invokePolicyError
-		if errors.As(err, &policyErr) && policyErr.cause != nil {
-			telemetry.RecordToolError(ctx, nil, serverConfig.Name, inferServerTransportType(serverConfig), req.Params.Name)
+		clientName := ""
+		if clientInfo := auditClientInfoFromSession(req.Session); clientInfo != nil {
+			clientName = clientInfo.Name
 		}
-		return result, err
+
+		startTime := time.Now()
+		serverType := inferServerTransportType(serverConfig)
+		spanAttrs := []attribute.KeyValue{
+			attribute.String("mcp.server.name", serverConfig.Name),
+			attribute.String("mcp.server.type", serverType),
+		}
+		if serverConfig.Spec.Image != "" {
+			spanAttrs = append(spanAttrs, attribute.String("mcp.server.image", serverConfig.Spec.Image))
+		}
+		if serverConfig.Spec.SSEEndpoint != "" {
+			spanAttrs = append(spanAttrs, attribute.String("mcp.server.endpoint", serverConfig.Spec.SSEEndpoint))
+		} else if serverConfig.Spec.Remote.URL != "" {
+			spanAttrs = append(spanAttrs, attribute.String("mcp.server.endpoint", serverConfig.Spec.Remote.URL))
+		}
+
+		ctx, span := telemetry.StartToolCallSpan(ctx, req.Params.Name, spanAttrs...)
+		metricAttrs := []attribute.KeyValue{
+			attribute.String("mcp.server.name", serverConfig.Name),
+			attribute.String("mcp.server.type", serverType),
+			attribute.String("mcp.tool.name", req.Params.Name),
+			attribute.String("mcp.client.name", clientName),
+		}
+		defer func() {
+			telemetry.ToolCallDuration.Record(ctx, float64(time.Since(startTime).Milliseconds()), metric.WithAttributes(metricAttrs...))
+			span.End()
+		}()
+		telemetry.ToolCallCounter.Add(ctx, 1, metric.WithAttributes(metricAttrs...))
+
+		result, err := next(ctx, req)
+		if err != nil {
+			telemetry.RecordToolError(ctx, span, serverConfig.Name, serverType, req.Params.Name)
+			span.SetStatus(codes.Error, "Tool call failed")
+			return result, err
+		}
+
+		span.SetStatus(codes.Ok, "")
+		return result, nil
 	}
 }
--- a/pkg/gateway/tool_policy.go
+++ b/pkg/gateway/tool_policy.go
@@ -1,0 +1,51 @@
+package gateway
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+
+	"github.com/docker/mcp-gateway/pkg/policy"
+)
+
+type invokePolicyError struct {
+	server string
+	tool   string
+	reason string
+	cause  error
+}
+
+func (e *invokePolicyError) Error() string {
+	if e.cause != nil {
+		return fmt.Sprintf("policy check failed for %s/%s: %v", e.server, e.tool, e.cause)
+	}
+	return fmt.Sprintf("policy denied tool %s on server %s: %s", e.tool, e.server, e.reason)
+}
+
+func (e *invokePolicyError) Unwrap() error {
+	return e.cause
+}
+
+// withInvokePolicy applies the fail-closed ActionInvoke gate to a backend tool
+// handler. Backend dispatch paths wrap handlers with this middleware when the
+// handler is registered, so direct calls, mcp-exec, and code-mode share the
+// same policy and audit behavior.
+func (g *Gateway) withInvokePolicy(serverName, toolName string, next mcp.ToolHandler) mcp.ToolHandler {
+	return func(ctx context.Context, req *mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+		if g.policyClient != nil {
+			policyReq := g.configuration.policyRequest(serverName, toolName, policy.ActionInvoke)
+			decision, err := g.policyClient.Evaluate(ctx, policyReq)
+			event := buildAuditEvent(policyReq, decision, err, auditClientInfoFromSession(req.Session))
+			submitAuditEvent(g.policyClient, event)
+			if err != nil {
+				return nil, &invokePolicyError{server: serverName, tool: toolName, cause: err}
+			}
+			if !decision.Allowed {
+				return nil, &invokePolicyError{server: serverName, tool: toolName, reason: decision.Reason}
+			}
+		}
+
+		return next(ctx, req)
+	}
+}
--- a/pkg/gateway/tool_policy.go
+++ b/pkg/gateway/tool_policy.go
@@ -2,11 +2,14 @@ package gateway
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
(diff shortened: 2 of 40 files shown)


pkg/gateway/capabilitites.go
```diff
@@ -177,7 +177,11 @@ func (g *Gateway) listCapabilities(ctx context.Context, serverNames []string, cl
 						capabilities.Tools = append(capabilities.Tools, ToolRegistration{
 							ServerName: serverConfig.Name,
 							Tool:       &prefixedTool,
-							Handler:    g.mcpServerToolHandler(serverConfig.Name, g.mcpServer, tool.Annotations, tool.Name),
+							Handler: g.withInvokePolicy(
+								serverConfig.Name,
+								tool.Name,
+								g.mcpServerToolHandler(serverConfig.Name, g.mcpServer, tool.Annotations, tool.Name),
+							),
```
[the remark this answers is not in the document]
kgprs: Addressed in 21ae1cb4: MCP-backed registrations now place a transport-aware telemetry adapter outside the shared policy middleware, restoring exactly one mcp.tool.errors metric when ActionInvoke evaluation fails before the backend handler runs. Added a regression test covering the metric value and server, transport, and tool attributes while confirming the backend is not called. Validation: go test ./pkg/gateway; make format; make test; make lint.

vendor/github.com/docker/mcp-gateway-oauth-helpers/discovery.go
mickael-docker, line 1: @kgprs Should we make this change in the upstream github.com/docker/mcp-gateway-oauth-helpers and update the module version here?

kgprs: Addressed in c382afe3. I moved the discovery hardening into upstream docker/mcp-gateway-oauth-helpers#9, updated this repo to the immutable upstream PR-head pseudo-version (16fe741), and regenerated vendor instead of maintaining a handwritten vendor-only patch. Upstream validation: make test, make lint, and CodeQL pass. Gateway validation: focused OAuth/gateway tests, helper compile, make format, make test, and make lint pass.

kgprs: Follow-up in 6a426b3d: removed the ci.yml changes from upstream #9, which now contains only discovery.go and discovery_test.go. The base workflow repair is isolated in docker/mcp-gateway-oauth-helpers#10. This PR now tracks the code-only upstream head fe5cca4 and keeps the regenerated vendor source unchanged. Focused OAuth/gateway tests and helper compilation pass.

kgprs: Final structure in d5176e2b: upstream #9 is stacked on the CI-only #10 branch. #9's visible diff contains only discovery.go and discovery_test.go, while inheriting the workflow prerequisite needed to run checks. This PR tracks upstream head 967ee4e; the vendored discovery source is unchanged and focused tests pass.

mickael-docker: Allow explicitly advertised cross-origin metadata. RFC 9728 does not require the resource_metadata URL to share the protected resource’s origin, so this rejects valid deployments that centrally host metadata on another HTTPS domain. The guarded HTTP client already blocks unsafe destinations, and validateProtectedResource subsequently binds the returned document to the requested resource. Please fetch the advertised URL through the guarded client without imposing same-origin.

mickael-docker: We should consider restoring the root metadata fallback. For a server such as https://example.com/mcp, this tries only /.well-known/oauth-protected-resource/mcp. MCP discovery also requires trying /.well-known/oauth-protected-resource at the origin root. A server publishing only the root document will now fail to reveal its authorization server, although the previous implementation supported it. Please try both locations before falling back to the resource origin.

mickael-docker (approved):
LGTM, thanks !

pkg/gateway/tool_policy.go
```diff
@@ -0,0 +1,69 @@
...
+// while POCI registrations use their existing outer telemetry middleware.
+func withMCPServerPolicyErrorTelemetry(serverConfig *catalog.ServerConfig, next mcp.ToolHandler) mcp.ToolHandler {
+	return func(ctx context.Context, req *mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+		result, err := next(ctx, req)
+		var policyErr *invokePolicyError
+		if errors.As(err, &policyErr) && policyErr.cause != nil {
+			telemetry.RecordToolError(ctx, nil, serverConfig.Name, inferServerTransportType(serverConfig), req.Params.Name)
+		}
+		return result, err
+	}
```
myang3577: `withMCPServerPolicyErrorTelemetry` only calls `telemetry.RecordToolError` when `policyErr.cause != nil` (evaluator failure), never for a plain `!decision.Allowed` denial. Since `withInvokePolicy` returns before calling `next` on a denial, `mcpServerToolHandler`'s own span/counter/duration never runs either — so a plain policy denial on the MCP-server dispatch path now produces no telemetry at all. This reproduces the gap reviewer kgprs flagged on the POCI path in #546 (fixed there by recording telemetry for any error, not just evaluator failures) — worth mirroring that fix here.

myang3577 (commented):
### Code review

Found 2 findings. One (the `tool_policy.go` telemetry gap) is posted as an inline comment on this PR's diff. The other (the `activate-profile` policy bypass) is included below instead, because it's in a file this PR doesn't touch and GitHub can't anchor an inline comment on a line outside the diff.

**mcp-activate-profile still lets a noop/missing policy activate catalog-only servers — the exact gap this PR claims to close**

This PR adds `checkServerManagementAccess`, which requires a real, non-noop policy provider before `mcp-add`/`mcp-config-set` can activate a catalog-only server. But `ActivateProfile` (`pkg/gateway/activateprofile.go:134`) still calls the old, unguarded `checkServerLoadPolicy` directly, which allows the action whenever the policy client is nil or the default no-op client — exactly the self-hosted configuration this PR is hardening against. `checkServerLoadPolicy`'s own pre-existing docstring (`configset.go:195-204`) names `activate-profile` as a caller this gate is meant to protect symmetrically with `mcp-config-set`, so this reads as an overlooked call site rather than an intentional difference.

https://github.com/docker/mcp-gateway/blob/5efe498ee4fedeca9eb3629852310696b438cf1b/pkg/gateway/activateprofile.go#L130-L138

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

pkg/gateway/tool_policy.go
kgprs: Addressed in 7d332bd6: moved the complete MCP-server telemetry envelope outside ActionInvoke, so both plain denials and evaluator failures record exactly one tool-call counter, duration, error counter, and error span without invoking the backend. The inner handler instrumentation was removed to avoid double-counting allowed calls. Validation: focused telemetry tests, focused race test, full pkg/gateway tests, go vet, make test, and make lint.

kgprs:
Also addressed the non-inline review-summary finding in 7d332bd6: mcp-activate-profile now uses the same enforcing-policy management boundary as mcp-add and mcp-config-set, while building ActionLoad requests from the profile's server metadata. Nil/noop policy is rejected before validation or mutation. Added nil/noop regression coverage and updated the security documentation. Validation: focused tests, focused race test, full pkg/gateway tests, go vet, make test, and make lint.

myang3577 (approved):

PR #584, opened 2026-09-16
fix: translate pre-2020-12 tool schema dialects to JSON Schema 2020-12

cutecatfann:
**What I did**

Tool calls to some catalog servers fail in every client that validates tool schemas with a JSON Schema 2020-12-only validator. The client rejects the tool before dispatch:

```
Tool 'list_bases' has an invalid outputSchema: JSON Schema declares an unsupported
dialect ("$schema": "http://json-schema.org/draft-07/schema#"). The default validator
supports JSON Schema 2020-12 only.
```

The gateway now translates tool `inputSchema` and `outputSchema` from JSON Schema draft-04/06/07 into 2020-12 before advertising them. `--preserve-tool-schema-dialect` restores the previous verbatim relaying.

**Why this is the gateway's problem**

Neither side is misbehaving, which is why the middle is the right place to fix it. MCP lets a schema declare any dialect, requires implementations to support only 2020-12, and tells them to reject an unsupported dialect with an error saying so — which is what these clients do. Servers built on the MCP TypeScript SDK declare draft-07 for *every* tool, because its zod converter defaults to that target. Both halves are conformant and the combination does not work.

So this is not specific to one server and not fixable by rebuilding an image: it affects any server whose schemas come from that SDK, and the upstream servers have nothing to fix. Reproduced directly against `@modelcontextprotocol/sdk` 1.24.3 with zod 4 — **both** `inputSchema` and `outputSchema` come back carrying the draft-07 declaration, so every tool of such a server is affected, not just the one the client happens to name.

**Why translate rather than strip or relabel `$schema`**

Because a bare relabel is unsound, and this is measured rather than argued. zod emits draft-07 tuple form for `z.tuple(...)` — `"items": [A, B]` plus `"additionalItems"` — and array-form `items` is not merely deprecated under 2020-12 but *invalid*: ajv's 2020-12 instance rejects the schema with `schema is invalid: data/items must be object,boolean`. Relabelling would trade a dialect error for a schema-invalid error. `dependencies` is worse: a 2020-12 validator ignores it as an unknown keyword, so the constraint is silently lost.

**What it refuses to do**

`pkg/toolschema` acts only on a schema that *explicitly* declares draft-04/06/07. A schema with no `$schema` is already 2020-12 by the spec's default and is left alone rather than guessed at.

Where no faithful equivalent exists, the schema is relayed exactly as declared and the reason logged — so the fallback is the behaviour that predates this PR, never a plausible-looking lie. Those cases are a `$ref` carrying an assertion sibling (one dialect ignores it, the other enforces it), a keyword 2020-12 introduced (inert where the schema was written, live once it claims 2020-12), a nested `$schema`, a plain-name-fragment `$id`, draft-04 carrying both `id` and `$id`, a draft-04 exclusive flag with no bound, and a non-object `dependencies`.

Deliberately **not** renaming `definitions` to `$defs`: measured against ajv's 2020-12 instance in strict mode, a `definitions` section and pointers into it resolve untouched, because `$ref` resolution is JSON Pointer over the raw document. The rename bought nothing and any rewrite of pointer text corrupts a pointer that legitimately traverses a property named `definitions`.

**Security**

This package is the first thing in the gateway to *walk* a tool schema rather than relay it opaquely, and a tool schema is untrusted input from across a trust boundary. The walk is bounded in depth and node count and leaves the schema untranslated beyond either limit; `docs/security.md` records it.

**Observability**

`mcp.tool.schema_dialects` counts schemas per server by outcome (`translated` / `relayed`), which is what makes the real blast radius measurable instead of estimated.

**Also in here**

`docs/mcp-gateway.md`'s hand-pasted `--help` dump is replaced by a link to the CI-validated generated reference. It named one flag that no longer exists, omitted fourteen that do, and had three wrong types — adding one line to it would have made it look maintained while staying wrong about eighteen other things.

**Testing**

- `go test ./pkg/toolschema/ ./pkg/telemetry/ ./pkg/gateway/`, `go vet`, `golangci-lint` clean on the changed packages.
- Every translation that has an oracle is pinned by a *verdict-equivalence* test rather than a shape assertion: the translated schema must accept and reject exactly what the original did. The two places the oracle cannot reach (draft-04, which `google/jsonschema-go` cannot parse; and post-2020-12 keywords, which it enforces regardless of the declared dialect) are called out in the test file and covered another way — the draft-04 bound rewrite is pinned on the boundary value, which is the only thing distinguishing it from its own inverse.
- Mutation-checked: a strip-only implementation, a relabel-only implementation, a removed bail-out and a removed walk budget are each killed by named tests.
- `make docs` output is committed and verified byte-identical by re-running the generator (the `validate-docs` CI stage checks this).

Not yet run: the end-to-end check against a live catalog image, which needs a Docker daemon.


austin5456:
The integration shape, bounded traversal, fallback behavior, and test coverage are thoughtful, but I do not think the default-on translator is safe to merge yet.

Blocking correctness gaps:

1. The unsupported-keyword guard is not source-dialect-aware. Draft-04 does not define `const`, `contains`, or `propertyNames`, and draft-06 does not define `if` / `then` / `else`. Those keywords are currently copied and become active after the document is relabelled 2020-12, changing what the source schema accepted. This contradicts the core guarantee that a translation is equivalent or relayed unchanged. A dialect-specific allowlist or equivalent source-keyword model would be safer than extending the current partial blacklist.
2. Identifier handling can change reference resolution or produce an invalid 2020-12 schema. Draft-04 `$id` is unknown but is copied through and becomes active. The fragment guard catches `#row` but misses non-empty fragments such as `#/row` and `https://example.test/schema#row`. Draft-04 numeric `exclusiveMinimum` / `exclusiveMaximum` has the same activation problem.
3. A root `$ref` candidate always falls back because its required root `$schema` declaration is treated as an unsafe sibling. Exempt the root declaration while retaining the nested-`$schema` guard.
4. `contentSchema` is another version-sensitive keyword absent from the bailout model. This is another reason to make the handling dialect-specific rather than relying on a shared blacklist.

Non-blocking cleanup: the public `Normalize` contract says returned containers never alias the input, but no-op and unsupported paths return the original value. Either deep-copy those paths or narrow the documentation.

All current checks pass, but none of the tests cover the source-dialect activation cases above. Since this PR enables translation by default, I would fix these before merge and add regression cases for each source dialect.

cutecatfann:
@austin5456 thanks, the dialect-awareness finding was the right call and the model was wrong rather than just incomplete. All four blocking items and the cleanup are fixed in `e85fe3d4`.

**1. Source-dialect-aware guard.** Replaced the blacklist with two sets: the keywords 2020-12 treats as significant (its core, applicator, validation and unevaluated vocabularies, plus `contentSchema`), and the keywords each draft actually defines. A keyword in the first and not the second is a refusal. Both sets are read from the published meta-schemas (`json-schema.org/draft-0N/schema` and the 2020-12 `meta/*` vocabularies) rather than written from memory, with one manual addition: draft-04's meta-schema omits `$ref`, which draft-04 does define, because `$ref` short-circuits there.

Annotations are deliberately excluded from the significant set, since they cannot change an outcome and including them would cost needless refusals. `$defs` is excluded for the same reason: 2020-12 defines it, but it only holds subschemas. This also dropped two refusals the blacklist was causing, as `$recursiveRef` and `$recursiveAnchor` are unknown to 2020-12 as well and so cannot activate.

**2. Identifiers.** Now refusing any `#` in an identifier, which covers `#/row` and `https://example.test/s#row` as well as the bare plain name. draft-04 `$id` needs no special case any more, it falls out of the model since draft-04 does not define it. A draft-04 numeric `exclusiveMinimum`/`exclusiveMaximum` is now refused rather than carried across. My earlier reasoning there was backwards: the form is right and the activation is the problem, and it is the same mistake as any other undefined keyword.

**3. Root `$ref`.** The root `$schema` declaration is exempt from the sibling rule, nested `$schema` is still refused. `definitions` and `$defs` are exempt too, since neither asserts anything in any dialect and a root `$ref` normally points into one. Without that second part the case still failed.

**4. `contentSchema`** is in the significant set, so it is refused for all three source dialects. Kept as the conservative call even though the content vocabulary is annotation-only by default, since a refusal costs only a relay.

**Cleanup:** narrowed the `Normalize` contract to what the code does. The translating path deep-copies; the no-op, unsupported-dialect and refusal paths return the input value itself, which is the point of those paths.

**Tests.** Added a per-source-dialect table covering each activation case, including the `const` row where relabelling turns "accepts every instance" into "accepts only x", and negative rows asserting an annotation does not cost a schema its translation. Also every identifier fragment form, and the root-`$ref` case. Mutation-checked: pinning the guard to a single dialect is killed by seven of those cases.

Two other things from an earlier internal pass, in case they affect how you read the diff. Translating `outputSchema` had no coverage at all (deleting the line left the suite green, since the assertion helper only read `inputSchema`), and a dialect outside draft-04/06/07 such as draft-03 returned a bare `Result{}`, indistinguishable from "nothing to do", so it now carries its own `unsupported_dialect` outcome. The fixture and server names are generalised.

Checks are still running on the new head. Happy to keep it default-off here instead if you would rather see it land that way first.

austin5456 (approved):

